PR #1, opened 2016-03-15
Add post-install/pre-removal scripts to link bins

jasonmp85:
These simply use `update-alternatives` to add alternative listings for the binaries packaged with Citus. If a system has Citus installed both for PostgreSQL 9.4 _and_ 9.5, then then 9.5 binaries have precedence unless the user manually changes their alternatives settings.

If only a single PostgreSQL version has Citus installed, then the alternatives system is technically extraneous, but results in all our binaries just showing up on the user's path.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -15,7 +15,20 @@ override_dh_auto_configure:
 	+pg_buildext configure build-%v
 
 override_dh_auto_install:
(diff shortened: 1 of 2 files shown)


jasonmp85:
Merging this.

PR #9, opened 2016-08-11
Register PGDG dependencies in install scripts

### LINKED ISSUE #8: Install fails on CentOS 7.2 / Fedora 24
Starting from a fresh machine this is what happens.

``` bash
$ curl https://install.citusdata.com/community/rpm.sh | sudo bash
# etc etc

$ sudo yum install citus_95
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: mirrors.evowise.com
 * extras: mirror.supremebytes.com
 * updates: mirror.supremebytes.com
Resolving Dependencies
--> Running transaction check
---> Package citus_95.x86_64 0:5.2.0-0.1.rc.1.el7.centos will be installed
--> Processing Dependency: postgresql95-server for package: citus_95-5.2.0-0.1.rc.1.el7.centos.x86_64
--> Processing Dependency: libpq.so.5()(64bit) for package: citus_95-5.2.0-0.1.rc.1.el7.centos.x86_64
--> Running transaction check
---> Package citus_95.x86_64 0:5.2.0-0.1.rc.1.el7.centos will be installed
--> Processing Dependency: postgresql95-server for package: citus_95-5.2.0-0.1.rc.1.el7.centos.x86_64
---> Package postgresql-libs.x86_64 0:9.2.15-1.el7_2 will be installed
--> Finished Dependency Resolution
Error: Package: citus_95-5.2.0-0.1.rc.1.el7.centos.x86_64 (citusdata_community)
           Requires: postgresql95-server
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
```

If I then try yum installing `postgresql95-server` it can't find the package.

---

Fedora 24 also fails to find that dependency:

``` bash
$ sudo yum install citus_95
Yum command has been deprecated, redirecting to '/usr/bin/dnf install citus_95'.
See 'man dnf' and 'man yum2dnf' for more information.
To transfer transaction metadata from yum to DNF, run:
'dnf install python-dnf-plugins-extras-migrate && dnf-2 migrate'

Failed to synchronize cache for repo 'citusdata_community-source', disabling.
Last metadata expiration check: 0:00:07 ago on Tue Aug  9 23:47:22 2016.
Error: nothing provides postgresql95-server needed by citus_95-5.1.1-1.fc24.x86_64
(try to add '--allowerasing' to command line to replace conflicting packages)
```

Another thing to note: Fedora appears to want to install 5.1 whereas CentOS is on 5.2 RC 1.

begriffs:
With this change the user will be able to immediately install postgresql-9.5-citus or citus_95 after running the Citus community installer for their Linux.

Fixes #8 

Notes
- Installers are still idempotent
- deb.sh now creates `/etc/apt/sources.list.d/pgdg.list` and will _overwrite_ it if it exists. Theoretically it might clobber a user file, but it seems unlikely
- Not sure what happens if the system had an older postgres version already installed, should I test this?
- RPM version has a check to deal with 32bit systems although IIRC we don't have a Citus package created for this yet. It's for future proofing.

@jasonmp85 could give these scripts one more test in your battery of Docker containers?

### DIFF
--- a/community/deb.sh
+++ b/community/deb.sh
@@ -8,14 +8,61 @@ unknown_os ()
   exit 1
 }
 
+arch_check ()
+{
+  if [ "$(uname -m)" != 'x86_64' ]; then
+    echo "Unfortunately, the Citus repository does not contain packages for non-x86_64 architectures."
+    echo
+    echo "Please email engage@citusdata.com with any issues."
+    exit 1
+  fi
+}
+
 curl_check ()
 {
   echo "Checking for curl..."
   if command -v curl > /dev/null; then
     echo "Detected curl..."
   else
-    echo "Installing curl..."
-    apt-get install -q -y curl
+    echo -n "Installing curl... "
+    apt-get install -y --no-install-recommends curl &> /dev/null
+    echo "done."
+  fi
+}
+
+pgdg_check ()
+{
+  echo "Checking for postgresql-9.5..."
+  if apt-cache show postgresql-9.5 &> /dev/null; then
+    echo "Detected postgresql-9.5..."
+  else
+    pgdg_list='/etc/apt/sources.list.d/pgdg.list'
+    pgdg_source_path="deb http://apt.postgresql.org/pub/repos/apt/ ${codename}-pgdg main"
+    pgdg_key_url='https://www.postgresql.org/media/keys/ACCC4CF8.asc'
+
+    if [ -e $pgdg_list ]; then
+      echo "Unable to install PostgreSQL Apt Repository"
+      echo
+      echo "The file ${pgdg_list} already exists."
+      echo
+      echo "Contact engage@citusdata.com with information about your system for help."
+      exit 1
+    fi
+
+    echo -n "Installing ${pgdg_list}... "
+
+    # create an apt config file for the PGDG repository
+    echo "${pgdg_source_path}" > $pgdg_list
+    echo "done."
+
+    echo -n "Installing ca-certificates... "
+    apt-get install -y --no-install-recommends ca-certificates &> /dev/null
+    echo "done."
+
+    echo -n "Importing PostgreSQL gpg key... "
+    # import the gpg key
+    curl -L "${pgdg_key_url}" 2> /dev/null | apt-key add - &>/dev/null
+    echo "done."
   fi
 }
 
@@ -79,16 +126,45 @@ detect_os ()
   echo "Detected operating system as $os/$dist."
 }
 
+detect_codename ()
+{
+  if [ "${os}" = "debian" ]; then
+    case "${dist}" in
+      7)
+        codename='wheezy'
+        ;;
+      8)
+        codename='jessie'
+        ;;
+      wheezy)
+        codename="${dist}"
+        ;;
+      jessie)
+        codename="${dist}"
+        ;;
+      *)
+        unknown_os
+        ;;
+    esac
+  else
+    codename=${dist}
+  fi
+}
+
 main ()
 {
   detect_os
-  curl_check
+  detect_codename
 
   # Need to first run apt-get update so that apt-transport-https can be
   # installed
   echo -n "Running apt-get update... "
   apt-get update &> /dev/null
   echo "done."
+
+  arch_check
+  curl_check
+  pgdg_check
 
   # Install the debian-archive-keyring package on debian systems so that
   # apt-transport-https can be installed next
@@ -104,7 +180,7 @@ main ()
 
   apt_source_path="/etc/apt/sources.list.d/citusdata_community.list"
 
-  echo -n "Installing $apt_source_path..."
+  echo -n "Installing $apt_source_path... "
 
   # create an apt config file for this repository
   curl -sSf "${apt_config_url}" > $apt_source_path
@@ -157,7 +233,7 @@ main ()
   echo "done."
 
   echo
-  echo "The repository is setup! You can now install packages."
+  echo "The repository is set up! You can now install packages."
 }
 
 main
--- a/community/rpm.sh
+++ b/community/rpm.sh
@@ -8,14 +8,38 @@ unknown_os ()
   exit 1
 }
 
+arch_check ()
+{
+  if [ "$(uname -m)" != 'x86_64' ]; then
+    echo "Unfortunately, the Citus repository does not contain packages for non-x86_64 architectures."
+    echo
+    echo "Please email engage@citusdata.com with any issues."
+    exit 1
+  fi
+}
+
 curl_check ()
 {
   echo "Checking for curl..."
   if command -v curl > /dev/null; then
     echo "Detected curl..."
   else
-    echo "Installing curl..."
+    echo -n "Installing curl... "
     yum install -d0 -e0 -y curl
+    echo "done."
+  fi
+}
+
+pgdg_check ()
+{
+  echo "Checking for postgresql95-server..."
+  if yum list -q postgresql95-server &> /dev/null; then
+    echo "Detected postgresql95-server..."
+  else
+    echo -n "Installing pgdg95 repo... "
+
+    yum install -d0 -e0 -y "${repo_url}"
+    echo "done."
   fi
 }
 
@@ -78,17 +102,64 @@ detect_os ()
   echo "Detected operating system as ${os}/${dist}."
 }
 
+detect_repo_url ()
+{
+  # set common defaults used by most flavors
+  family='redhat'
+  family_short='rhel'
+  pkg_dist="${dist}"
+  pkg_os="${os}"
+  pkg_version='2'
+
+  case "${os}" in
+    amzn)
+      # require at least a 2015 image
+      if [ "${dist}" -lt "2015" ]; then
+        unknown_os
+      fi
+
+      # use 2015.03 pgdg repo for all recent Amazon instances
+      pkg_dist=6
+      pkg_os='ami201503-'
+      ;;
+    ol)
+      pkg_os='oraclelinux'
+      ;;
+    fedora)
+      family='fedora'
+      family_short='fedora'
+      pkg_version='3'
+      ;;
+    centos)
+      # defaults are suitable
+      ;;
+    rhel|redhatenterpriseserver)
+      pkg_os='redhat'
+      ;;
+    *)
+      unknown_os
+      ;;
+  esac
+
+  repo_url="https://download.postgresql.org/pub/repos/yum/9.5/${family}"
+  repo_url+="/${family_short}-${pkg_dist}-x86_64"
+  repo_url+="/pgdg-${pkg_os}95-9.5-${pkg_version}.noarch.rpm"
+}
+
 main ()
 {
   detect_os
-  curl_check
+  detect_repo_url
 
+  arch_check
+  curl_check
+  pgdg_check
 
   yum_repo_config_url="https://repos.citusdata.com/community/config_file.repo?os=${os}&dist=${dist}&source=script"
 
   yum_repo_path=/etc/yum.repos.d/citusdata_community.repo
 
-  echo "Downloading repository file: ${yum_repo_config_url}"
+  echo -n "Downloading repository file: ${yum_repo_config_url}... "
 
   curl -sSf "${yum_repo_config_url}" > $yum_repo_path
   curl_exit_code=$?
@@ -131,8 +202,8 @@ main ()
     echo "done."
   fi
 
-  echo "Installing pygpgme to verify GPG signatures..."
-  yum install -y pygpgme --disablerepo='citusdata_community'
+  echo -n "Installing pygpgme to verify GPG signatures... "
+  yum install -d0 -e0 -y pygpgme --disablerepo='citusdata_community' &> /dev/null
   pypgpme_check=`rpm -qa | grep -qw pygpgme`
   if [ "$?" != "0" ]; then
     echo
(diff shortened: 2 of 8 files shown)


lfittl:
@begriffs Re: `Not sure what happens if the system had an older postgres version already installed, should I test this?` - this seems like it could potentially be a problem if someone has an existing PostgreSQL cluster that they want to add the citus extension too.

My assumption is that this might force their hand into an upgrade to PostgreSQL 9.5 - which is needed for Citus, but might not be obvious from the instructions. Can we add a easily noticable warning that we're doing this? (auto-adding PGDG)

community/deb.sh
```diff
@@ -79,10 +79,19 @@ detect_os ()
   echo "Detected operating system as $os/$dist."
 }
 
+add_pgdg ()
+{
+  echo "Registering PostgreSQL PGDG dependencies..."
+  echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list
+  apt-get install wget ca-certificates
+  wget --quiet --no-check-certificate -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
```
jasonmp85: Eeeehhh, a dependency on _both_ `curl` _and_ `wget`? We should change this to `curl`.

begriffs: Good catch, I lifted that from a snippet I had in the docs.

community/deb.sh
```diff
@@ -79,10 +79,19 @@ detect_os ()
   echo "Detected operating system as $os/$dist."
 }
 
+add_pgdg ()
+{
+  echo "Registering PostgreSQL PGDG dependencies..."
+  echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list
```
jasonmp85: I feel a _little_ uneasy blasting over this file without checking first. I know this script also blasts over the packagecloud file, but we "own" that one. This one could already be there on a user's machine.

community/deb.sh
```diff
@@ -79,10 +79,19 @@ detect_os ()
   echo "Detected operating system as $os/$dist."
 }
 
+add_pgdg ()
```
jasonmp85: I think I'd like to verify whether `postgresql-9.5` is available before adding the repository. Looking into how to do this…

jasonmp85: Ok, in Debian:

``` sh
apt-cache show postgresql-9.5 2> /dev/null
```

This prints stuff about the available package to stdout. Can capture to a variable and test for empty string, or use exit code.

community/rpm.sh
```diff
@@ -78,10 +78,39 @@ detect_os ()
   echo "Detected operating system as ${os}/${dist}."
 }
 
+add_pgdg ()
```
jasonmp85: And checking this in Red Hat is:

``` sh
yum list -q postgresql95-server 2>/dev/null
```

Again, can check for empty string or use exit code.

community/rpm.sh
```diff
@@ -78,10 +78,39 @@ detect_os ()
...
+  fi
+
+  if [[ "$(arch)" = "x86_64" ]]; then
+    arch="x86_64"
+  else
+    # sometimes reports as i686
+    arch="i386"
+  fi
+
+  yum install -y https://download.postgresql.org/pub/repos/yum/9.5/$family/$family_short-$dist-$arch/pgdg-${pkg_os}95-9.5-$suffix.noarch.rpm
```
jasonmp85: This is _super_ long. Maybe break across several lines, building it up via appends?

community/deb.sh
jasonmp85: Also, out of analogy to `curl_check`, I think we should call this `pgdg_check`.

community/deb.sh
```diff
@@ -79,10 +79,19 @@ detect_os ()
...
+  echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list
+  apt-get install ca-certificates
+  curl -L https://www.postgresql.org/media/keys/ACCC4CF8.asc 2> /dev/null | apt-key add - &>/dev/null
+}
+
 main ()
 {
   detect_os
   curl_check
+  add_pgdg
```
jasonmp85: Since I'd like to add a `postgresql-9.5` availability check in here, let's move this after the _Running apt-get update_ phase.

jasonmp85: Here are the three scenarios I envision, as seen by the user:

# Debian

## PostgreSQL 9.5 Found

```
Checking for postgresql-9.5...
Detected postgresql-9.5...
# no further PGDG-related output
```

## PostgreSQL 9.5 Not Found, `pgdg.list` File Absent

```
Checking for postgresql-9.5...
Installing PostgreSQL Apt Repository (PGDG)...
# no further PGDG-related output
```

## PostgreSQL 9.5 Not Found, `pgdg.list` File Present

```
Checking for postgresql-9.5...
Unable to install PostgreSQL Apt Repository

The file /etc/apt/sources.list.d/pgdg.list already exists.

Contact engage@citusdata.com with information about your system for help.
# exit with status 1
```

# Red Hat

## PostgreSQL 9.5 Found

```
Checking for postgresql95-server...
Detected postgresql95-server...
# no further PGDG-related output
```

## PostgreSQL 9.5 Not Found

```
Checking for postgresql95-server...
Installing PostgreSQL Yum Repository (PGDG)...
# no further PGDG-related output
```

community/deb.sh
```diff
@@ -79,17 +79,47 @@ detect_os ()
...
+The file $pgdg_list already exists.
+
+Contact engage@citusdata.com with information about your system for help.
+
+EOM
+      exit 1
+    fi
+
+    echo "Installing PostgreSQL Apt Repository (PGDG)..."
+    echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" > $pgdg_list
```
begriffs: Ran into an awkward thing. If the `lsb-release` package is not installed prior to running this script then the `$dist` variable is set to 8, otherwise to `jessie`. We need the codename when running the deb command. One solution is to install the package prior to running `detect_os`...

sumedhpathak:
I tried just running the rpm.sh script on Amazon Linux. It doesn't appear to successfully install the PGDG repository. 

Possibly relevant error:

```
Loaded plugins: priorities, update-motd, upgrade-helper
Cannot open: https://download.postgresql.org/pub/repos/yum/9.5/redhat/rhel-2016-x86_64/pgdg-amzn95-9.5-2.noarch.rpm. Skipping.
Error: Nothing to do
Downloading repository file: https://repos.citusdata.com/community/config_file.repo?os=amzn&dist=2016&source=script
done.
```

I also get this error at the bottom:

```
Importing GPG key 0x1530DF18:
 Userid     : "https://packagecloud.io/citusdata/community (https://packagecloud.io/docs#gpg_signing) <support@packagecloud.io>"
 Fingerprint: e6ff c1a8 e256 1fc0 04d1 847f 56f4 fa44 1530 df18
 From       : https://repos.citusdata.com/community/gpgkey
https://repos.citusdata.com/community/el/6/x86_64/repodata/repomd.xml: [Errno -1] repomd.xml signature could not be verified for citusdata_community
Trying other mirror.


 One of the configured repositories failed (citusdata_community),
 and yum doesn't have enough cached data to continue. At this point the only
 safe thing yum can do is fail. There are a few ways to work "fix" this:

     1. Contact the upstream for the repository and get them to fix the problem.

     2. Reconfigure the baseurl/etc. for the repository, to point to a working
        upstream. This is most often useful if you are using a newer
        distribution release than is supported by the repository (and the
        packages for the previous distribution release still work).

     3. Disable the repository, so yum won't use it by default. Yum will then
        just ignore the repository until you permanently enable it again or use
        --enablerepo for temporary usage:

            yum-config-manager --disable citusdata_community

     4. Configure the failing repository to be skipped, if it is unavailable.
        Note that yum will try to contact the repo. when it runs most commands,
        so will have to try and fail each time (and thus. yum will be be much
        slower). If it is a very temporary problem though, this is often a nice
        compromise:

            yum-config-manager --save --setopt=citusdata_community.skip_if_unavailable=true

failure: repodata/repomd.xml from citusdata_community: [Errno 256] No more mirrors to try.
```

However, it does seem to have installed our packageCloud repo. When I try install citus_95 however, it fails since PGDG itself wasn't set up:

```
[ec2-user@ip-10-142-209-58 ~]$ sudo yum install citus_95
Loaded plugins: priorities, update-motd, upgrade-helper
Resolving Dependencies
--> Running transaction check
---> Package citus_95.x86_64 0:5.2.0-0.1.rc.1.el6 will be installed
--> Processing Dependency: postgresql95-server for package: citus_95-5.2.0-0.1.rc.1.el6.x86_64
--> Processing Dependency: libpq.so.5()(64bit) for package: citus_95-5.2.0-0.1.rc.1.el6.x86_64
--> Running transaction check
---> Package citus_95.x86_64 0:5.2.0-0.1.rc.1.el6 will be installed
--> Processing Dependency: postgresql95-server for package: citus_95-5.2.0-0.1.rc.1.el6.x86_64
---> Package postgresql8-libs.x86_64 0:8.4.20-5.52.amzn1 will be installed
--> Finished Dependency Resolution
Error: Package: citus_95-5.2.0-0.1.rc.1.el6.x86_64 (citusdata_community)
           Requires: postgresql95-server
```

If I manually first install PGDG via our current documentation instructions, then run rpm.sh, I am able to update Citus 5.1 to Citus 5.2-rc1 package. This works fine.

sumedhpathak:
We generate the PGDG repo string as:

```
https://download.postgresql.org/pub/repos/yum/9.5/redhat/rhel-2016-x86_64/pgdg-amzn95-9.5-2.noarch.rpm
```

However, the PostgreSQL installation page has the repo URL as:

```
https://download.postgresql.org/pub/repos/yum/9.5/redhat/rhel-6-x86_64/pgdg-ami201503-95-9.5-2.noarch.rpm
```

sumedhpathak:
@begriffs and I have been testing the script. I've tested so far on:
Ubuntu: 12.04, 14.04 and 16.04. I also tested on Amazon Linux. I'm able to:
- Fresh install Citus 5.2 and PostgreSQL 9.5.4 via the script.
- Install PostgreSQL 9.5.4 and Citus 5.1 via PGDG, run the script, and then upgrade to Citus 5.2 using our minor version upgrade instructions.
- Install PostgreSQL 9.4 and Citus 5.1 in parallel with PostgreSQL 9.5 and Citus 5.2.

One note: On Amazon Linux we don't seem to have citus_94 packages available. However they do exist in the Ubuntu PGDG repo's. Joe mentioned the same for CentOS and RHEL as well. Any reason why we don't have the 9.4 supported packages in the RHEL repo's?

begriffs:
I successfully tested a fresh install on
- CentOS 6 & 7
- Oracle Linux 7
- Debian Jessie & Wheezy

I also tested first installing `postgresql-9.4-citus` on the Debians and then running the packagecloud script and installing `postgresql-9.5-citus`. This worked and I was able to run commands from the _Distributed Tables → Hash Distribution_ section of the docs on the cluster. Both versions of Postgres coexist in different directories.

begriffs:
Had a problem when running the script on RHEL 7:

```
Detected operating system as rhel/7.
Checking for curl...
Detected curl...
Checking for postgresql95-server...
Installing pgdg95 repo... Cannot open: https://download.postgresql.org/pub/repos/yum/9.5/redhat/rhel-7-x86_64/pgdg-rhel95-9.5-2.noarch.rpm. Skipping.
Error: Nothing to do
done.
Downloading repository file: https://repos.citusdata.com/community/config_file.repo?os=rhel&dist=7&source=script... done.
Installing pygpgme to verify GPG signatures... done.
Installing yum-utils... done.
Generating yum cache for citusdata_community... done.

The repository is set up! You can now install packages.
```

Looks like this would fix it

``` diff
-https://download.postgresql.org/pub/repos/yum/9.5/redhat/rhel-7-x86_64/pgdg-rhel95-9.5-2.noarch.rpm
+https://download.postgresql.org/pub/repos/yum/9.5/redhat/rhel-7-x86_64/pgdg-redhat95-9.5-2.noarch.rpm
```

I'm experimenting with changing `$pkg_os` in the rhel case and will add my commit if it works.

jasonmp85:
> Looks like this would fix it

Ah, a hasty refactoring on my part. I started out with a very explicit `switch` block but later removed what looked redundant among all the cases. Should be a quick fix.

begriffs:
That fixed rhel 7!

On rhel 6 I get:

```
Detected operating system as redhatenterpriseserver/6.
Unfortunately, your operating system distribution and version are not supported by this script.
```

Also it's using the lsb_release conditional branch to detect that:

``` bash
$ lsb_release -i
Distributor ID: RedHatEnterpriseServer
```

On rhel 7 it uses os-release

``` bash
$ . /etc/os-release
$ echo $ID
rhel
```

begriffs:
Tested that this also works on Amazon Linux `2015.03`, `2015.09`, and `2016.03`

jasonmp85:
Is there anything else to be done here, or are we close to ready?

jasonmp85:
@sumedhpathak, @begriffs, is this a :shipit:?

If so, I'll squash the commits and make sure the other files have the changes we've put into the community one.

sumedhpathak:
So it now works on RHEL as well? If so, I think its a 🚢

sumedhpathak:
Could you let us know once its merged and deployed (?) to install.citusdata.com?

begriffs:
Looks good to me as well. Once it's merged I'll test the new installation docs and merge that too.

PR #12, opened 2016-09-21
Clean up enterprise install scripts

jasonmp85:
Handles a few issues:
- Renames `unique_id` to `CITUS_REPO_HOST_ID` to be clearer about what it's used for
- Ensures master token will not be echoed, passing it as a user parameter instead
- Ensures master tokens containing colons are handled (unlikely, but fixed anyhow)
- Allows host IDs which contain spaces, other non-URL-safe characters

Everything should be good to go re: enterprise install scripts after this is merged.

### DIFF
--- a/enterprise-nightlies/deb.sh
+++ b/enterprise-nightlies/deb.sh
@@ -216,11 +216,14 @@ main ()
     exit 1
   fi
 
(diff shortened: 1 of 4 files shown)


saicitus (approved):
Looks Good!

PR #13, opened 2016-10-27
Bump expected PostgreSQL version to 9.6

jasonmp85:
This also (implicitly) removes support for the sunset Fedora 22 distro, as PostgreSQL itself no longer supports that distro. As it was the only
distro with a package release other than '3', that block was simplified somewhat by the change.

### DIFF
--- a/community-nightlies/deb.sh
+++ b/community-nightlies/deb.sh
@@ -32,9 +32,9 @@ curl_check ()
 
 pgdg_check ()
 {
-  echo "Checking for postgresql-9.5..."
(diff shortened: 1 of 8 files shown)


jasonmp85:
Putting this up to get the nice diff view, but it cannot be merged until we release 6.0. I plan to split this into two reviews: we can immediately bump the nightly PostgreSQL to 9.6, but need to wait to bump the community repo to 9.6.

jasonmp85:
Closing in favor of #14, #15, and #16.

PR #29, opened 2017-02-09
Switch to 'fancy' versioning for Citus RPMs

jasonmp85:
I'm not 100% sure whether a self-conflicting package is best in the RPM world, but despite the warning everything seems to work well.

Merging this will release `citus60_95` and `citus60_96`. The version will be `6.0.1.citus-2` and existing Citus users will be auto-upgraded. This and all subsequent 'fancy' version packages will conflict with one another, ensuring two versions cannot be installed at the same time.


byucesoy (approved):
Looks good. I also tested it in my local. It only produces these messages but I think it is OK.

      1 citus_95.x86_64: E: useless-provides citus_95
      1 citus_96.x86_64: E: useless-provides citus_96
     20 configure: WARNING: unrecognized options: --disable-dependency-tracking

PR #32, opened 2017-02-09
Bump Redhat Citus to 6.1.0

byucesoy:
Produces 6.1.0 release.


citus.spec
```diff
@@ -7,12 +7,11 @@ Summary:	PostgreSQL-based distributed RDBMS
 Name:		%{sname}%{?pkginfix}_%{pgmajorversion}
 Provides:	%{sname}_%{pgmajorversion}
 Conflicts:	%{sname}_%{pgmajorversion}
-Obsoletes:	%{sname}_%{pgmajorversion} <= 6.0.1.citus-1%{dist}
-Version:	6.0.1.citus
+Version:	6.1.0.citus
 Release:	2%{dist}
```
jasonmp85: This is usually set to 1… I only set to 2 for the transitional packages; we should change it back unless we need to perform multiple builds for the same tag.

PR #33, opened 2017-02-10
Switch to 'fancy' versioning for Citus Enterprise RPMs

byucesoy:
Similar to community editions, we are switching to 'fancy' versioning
for Citus Enterprise.

Merging this will release citus-enterprise60_95 and citus-enterprise60_96.
The version will be 6.0.1.citus-2 and existing Citus users will be auto-upgraded.
This and all subsequent 'fancy' version packages will conflict with one
another, ensuring two versions cannot be installed at the same time.


jasonmp85:
I didn't realize this until today, but these packages already have a `Conflicts` field to prevent installation atop OSS Citus packages. So we want to make sure they Provide and Conflict with `citus_96`, which used to be a real package name but is now a virtual package/capability name that lets us ensure only one thing that provides "citus" can do so for a given PostgreSQL major version at a time.

I tested this with a test repo; worked fine.

PR #36, opened 2017-02-14
Switch to fancy versioning for Redhat Shard Rebalancer

byucesoy:


citus-rebalancer.spec
```diff
@@ -47,6 +50,9 @@ PATH=%{pginstdir}/bin:$PATH
 %{pginstdir}/share/extension/shard_rebalancer.control
 
 %changelog
+* Thu Nov 10 2016 - Burak <burak@citusdata.com> 6.0.0.citus-2
```
jasonmp85, line 53: Hm; just caught this. Probably should have used a recent date.

PR #41, opened 2017-05-10
Bump Redhat Citus Enterprise to 6.1.1

metdos:
Produces 6.1.1 release.


metdos:
Hey @jasonmp85, could review this? I think this is what Cloud needs to update their clusters.

PR #48, opened 2017-05-24
Bump Redhat Rebalancer to 6.2.0

jasonmp85:
Produces 6.2.0 release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ env:
     - TARGET_PLATFORM=ol/7
     - TARGET_PLATFORM=ol/6
 before_install:
-  - git clone -b v0.5.2 --depth 1 https://github.com/citusdata/tools.git
+  - git clone -b curl_bug --depth 1 https://github.com/citusdata/tools.git
   - sudo make -C tools install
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -19,7 +19,7 @@ env:
     - TARGET_PLATFORM=ol/7
     - TARGET_PLATFORM=ol/6
 before_install:
-  - git clone -b v0.5.2 --depth 1 https://github.com/citusdata/tools.git
```
fdr, line 22: `curl_bug`?

fdr (approved):

.travis.yml
jasonmp85: packagecloud returns a `400` status from their List Versions API for packages that don't yet have any releases. Since every major release now has a new package name (with the suffixes and baked-in version and whatnot) I realized this bit of code was wrong (I was using `curl`'s `-f` option).

I quickly hacked around the fix but need to productize it to interpret `400` as an empty list of versions rather than an error (right now, all errors are accepted). After I fix I'll release a new tag, but until then, just pointing at my branch in the travis file.

PR #49, opened 2017-05-24
Bump Citus Debian to 6.2.1

jasonmp85:
Also stripped off the prefix in the lintian overrides as it turns out
they are unnecessary.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+citus (6.2.1.citus-1) stable; urgency=low
+
+  * Relaxes version-check logic to avoid breaking non-distributed commands
+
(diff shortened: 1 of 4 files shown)


sumedhpathak:
Looks good, but I have no idea what the lintian overrides are/were. 😛

jasonmp85:
> but I have no idea what the lintian overrides are/were

`lintian` is a linter/style-checker for warnings and errors about Debian package builds. We had hardcoded the package name in the file, which is apparently unnecessary, we can just say "ignore this warning for any package being built".

For posterity, the documentation about the grammar for the lintian overrides file lives [here](https://lintian.debian.org/manual/section-2.4.html#section-2.4.1).

PR #51, opened 2017-05-24
Bump Citus Enterprise Debian to 6.2.1

jasonmp85:
Also stripped off the prefix in the lintian overrides as it turns out
they are unnecessary.


sumedhpathak:
Looks good, apart from my ignorance of the lintian one again. 🚢

PR #54, opened 2017-06-07
Bump Citus Debian to 6.2.2

velioglu:
Changes to bump Citus Debian version to 6.2.2


jasonmp85 (commented):
Looks good to me. Tested locally. Merging  to trigger release.

PR #61, opened 2017-07-13
Bump Debian Citus Enterprise to 6.2.3

byucesoy:


jasonmp85:
You had the weekday as Tuesday when the 13th is a Thursday, so that generated some warnings. Fixed, amended, and pushed.

jasonmp85 (approved):

PR #67, opened 2017-08-29
Add gnupg2 to Debian builds

byucesoy:
Docker image of Debian Stretch does not come with gnupg, which is necessary for our operations. With this change we add it to our debian based Dockerfiles.

We are also removing the command which installs postgresql-9.5-citus-6.0 and postgresql-9.6-citus-6.0 packages to get headers. Because there is no citus packages for Debian Stretch yet thus, that command fails. Also it looks like everything works as expected even without that command. @jasonmp85 Did I miss anything about this?

### DIFF
--- a/dockerfiles/debian-jessie-all/Dockerfile
+++ b/dockerfiles/debian-jessie-all/Dockerfile
@@ -2,7 +2,11 @@
 FROM debian:jessie
 
 # install build tools and PostgreSQL development files
-RUN apt-key adv --keyserver ha.pool.sks-keyservers.net --recv-keys B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8 \
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        dirmngr \
+        gnupg2 \
+    && apt-key adv --keyserver ha.pool.sks-keyservers.net --recv-keys B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8 \
     && echo 'deb http://apt.postgresql.org/pub/repos/apt/ jessie-pgdg main' > /etc/apt/sources.list.d/pgdg.list \
     && apt-get update \
     && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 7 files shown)


dockerfiles/debian-jessie-all/Dockerfile
```diff
@@ -2,7 +2,11 @@
 FROM debian:jessie
 
 # install build tools and PostgreSQL development files
-RUN apt-key adv --keyserver ha.pool.sks-keyservers.net --recv-keys B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8 \
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
```
jasonmp85, line 6: Is there a reason these can't be added to the `apt-get install` step below (which already has a long list of alphabetized package names?)

dockerfiles/debian-stretch-all/Dockerfile
```diff
@@ -32,11 +36,6 @@ RUN curl -sL https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64
          -o /usr/bin/jq \
     && chmod +x /usr/bin/jq
 
-# install latest Citus release to get header files
-RUN curl -s https://packagecloud.io/install/repositories/citusdata/community/script.deb.sh | bash \
```
jasonmp85, line 36: This is actually necessary for the shard rebalancer, as it relies on headers exposed by the community package (having it rely on the headers from community was easier than shoehorning an enterprise credential into our Travis build logic).

Is there a strong reason to remove this before we add Debian Stretch? That's a separate piece of work (i.e. we can do an OS audit and make sure we're not missing anything).

jasonmp85:
What here was necessary for packaging? Do the existing images in Docker Hub not build Citus 7.0 for some reason?

jasonmp85:
Closing in lieu of #70.

dockerfiles/debian-jessie-all/Dockerfile
byucesoy: It was there because they are needed for apt-key command.

dockerfiles/debian-stretch-all/Dockerfile
byucesoy: There is no reason to remove them before adding Debian Stretch.

At #70 we only removed that lines for Debian Stretch and since now we add Debian Stretch it is also OK to add this back in future.

byucesoy:
All responses above and this comment is for documenting purposes.

This was not mandatory for packaging. It was nice to have and now #70  and #71 does provides same(and more) functinality. It is OK to close this.

jasonmp85:
Hey @byucesoy I'm writing up what's needed for adding new OSes so we can do so more smoothly in the future. I think this is the first time it's come up in a release.

PR #96, opened 2017-10-06
Update install scripts with latest upstream changes

jasonmp85:
Some minor refactorings, better treatment of curl exit codes, and new OS additions necessitate a re-sync with upstream.

### DIFF
--- a/community-nightlies/deb.sh
+++ b/community-nightlies/deb.sh
@@ -136,10 +136,16 @@ detect_codename ()
       8)
         codename='jessie'
         ;;
+      9)
+        codename='stretch'
+        ;;
       wheezy)
         codename="${dist}"
         ;;
       jessie)
+        codename="${dist}"
+        ;;
+      stretch)
         codename="${dist}"
         ;;
       *)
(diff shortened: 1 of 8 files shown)


community-nightlies/deb.sh
```diff
@@ -198,7 +204,7 @@ main ()
     echo "If you are running a supported OS, please contact us via https://www.citusdata.com/about/contact_us and report this."
     [ -e $apt_source_path ] && rm $apt_source_path
     exit 1
-  elif [ "$curl_exit_code" = "35" ]; then
+  elif [ "$curl_exit_code" = "35" -o "$curl_exit_code" = "60" ]; then
```
begriffs, line 207: The shellcheck linter suggests:

[SC2166](https://github.com/koalaman/shellcheck/wiki/SC2166): Prefer [ p ] || [ q ] as [ p -o q ] is not well defined.

Also there are several [libcurl error codes](https://curl.haxx.se/libcurl/c/libcurl-errors.html) that have to do with cert stuff. I don't know enough about it to know which codes are relevant here, but curious about your motivation to add this particular one.

jasonmp85: All of the changes here came from the upstream packagecloud `deb.sh` and `rpm.sh` scripts, including the curl exit code checks. I don't know what's causing them to check for it, but I've been keeping our script as a superset of theirs so I just merged their changes into our scripts.

begriffs:
I copied deb.sh and rpm.sh from the community directory on this branch and verified that I was able use them to complete the [single-machine cluster install](https://docs.citusdata.com/en/v7.0/installation/development.html) instructions on Debian 9.2 x64 and Fedora 26 x64 respectively.

Also the changes on community/enterprise ±nightlies look consistent when I eyeball them. The PR looks good to me as long as the diff of these scripts against the packagecloud originals contains the things you expect.

PR #107, opened 2017-11-15
Switch to PG10 packages in our install scripts

### LINKED ISSUE #103: Update install scripts to detect PostgreSQL 10
PostgreSQL 10 will be the "favored" version of PostgreSQL as of Citus 7.1, so our scripts should assume users want it. The eight `deb`/`rpm` scripts in the `gh-pages` branch need to be updated (search for any occurrences of `9.6` or `96` and update appropriately).

byucesoy:
Fixes #103

### DIFF
--- a/enterprise/rpm.sh
+++ b/enterprise/rpm.sh
@@ -32,11 +32,11 @@ curl_check ()
 
 pgdg_check ()
 {
-  echo "Checking for postgresql96-server..."
-  if yum list -q postgresql96-server &> /dev/null; then
-    echo "Detected postgresql96-server..."
+  echo "Checking for postgresql10-server..."
+  if yum list -q postgresql10-server &> /dev/null; then
+    echo "Detected postgresql10-server..."
   else
-    echo -n "Installing pgdg96 repo... "
+    echo -n "Installing pgdg10 repo... "
 
     yum install -d0 -e0 -y "${repo_url}"
     echo "done."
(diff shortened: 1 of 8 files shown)


jasonmp85:
@sumedhpathak ready for a quick look on that second commit. I'm testing on each OS using ``docker run -v`pwd`/community:/community -it --rm <OS> /community/rpm.sh``, i.e. mounting the community directory within a docker container and then running the script. I've done CentOS 6/7, Fedora 25/26 and Oracle Linux 6/7 now. Everything looks good (I haven't tested Red Hat Enterprise Linux, since I think I'd have to spin up an EC2 instance for that).

jasonmp85:
I was wrong about Fedora. I tried 25 and 26; 27 just came out yesterday and doesn't work, of course.

enterprise/rpm.sh
```diff
@@ -186,26 +186,16 @@ detect_repo_url ()
...
-      pkg_os='ami201503-'
-      pkg_version='2'
-      ;;
     ol)
       pkg_os='oraclelinux'
       ;;
     fedora)
       family='fedora'
       family_short='fedora'
+      pkg_version='3'
```
sumedhpathak, line 198: Fedora still seems to be at package version 2. Is this not yet updated on their website?

jasonmp85: Yeah, and many of the others still say `1` on their website.

PR #108, opened 2017-11-15
Bump Debian Shard Rebalancer to 7.1.0

velioglu:


byucesoy (changes requested):
Please also modify Depends line in debian/control.in

byucesoy (approved):
:ship:

PR #115, opened 2017-12-08
Add rpm96 files

jasonmp85:


sumedhpathak (approved):
Tested the community version, which looks good.

PR #183, opened 2018-05-17
Bump enterprise debian Version

velioglu:


debian/changelog
```diff
@@ -1,3 +1,9 @@
+citus-enterprise (7.2.2.citus-1) stable; urgency=low
```
jasonmp85, line 1: This should be added _in order_ with the other versions, unfortunately… It's causing the version number in the nightlies repo to be messed up.

PR #198, opened 2018-07-25
redhat fix Turkish char in changelog

furkansahin:


metdos (approved):

jasonmp85:
Wait how is this a "fix"? Did this break something?

furkansahin:
It is not supposed to. But I did it in case.

jasonmp85:
> It is not supposed to. But I did it in case.

It should be fine. The original Citus packager is Turkish and has similar characters in his name so I already had to make sure the encodings and locales worked correctly.

PR #266, opened 2018-12-20
Bump debian topn to 2.2.1

hanefi:


hanefi:
I needed to use a newer `citus_package` to be able to support releasing packages for `ubuntu/bionic`

I believe that it is safe to do this, but I'd like to know if you approve this change or not.

velioglu (approved):

PR #268, opened 2018-12-21
Bump Redhat Shard Rebalancer to 8.1.0

hanefi:


pkgvars
```diff
@@ -1,6 +1,6 @@
 pkgname=citus-rebalancer
 hubproj=shard_rebalancer
 pkgdesc='Shard Rebalancer'
-pkglatest=8.0.0.citus-1
+pkglatest=8.1.0.citus-1
 releasepg=11
```
velioglu: Update them to 10,11

citus-rebalancer.spec
```diff
@@ -49,6 +49,9 @@ PATH=%{pginstdir}/bin:$PATH
 %{pginstdir}/share/extension/shard_rebalancer.control
 
 %changelog
+* Fri Dec 21 2018 - Hanefi Onaldi <hanefi@citusdata.com> 8.1.0.citus-1
+- Official release for 8.1.0
+
 * Tue Nov 06 2018 - Burak Velioglu <velioglub@citusdata.com> 8.0.0.citus-1
 - Official release for 8.0.0
```
velioglu, line 67: You should add steps similar to https://github.com/citusdata/packaging/commit/7589e5907ba67de8899e50f33005b203a509a484

velioglu (changes requested):

pkgvars
hanefi: I changed this line to `releasepg=10,11`

citus-rebalancer.spec
hanefi: I added that part but I could not confirm if it is working properly now

pkgvars
hanefi: I also checked that the travis job builds packages for both pg versions. So it's ok

citus-rebalancer.spec
```diff
@@ -47,8 +47,21 @@ PATH=%{pginstdir}/bin:$PATH
 %{pginstdir}/lib/shard_rebalancer.so
 %{pginstdir}/share/extension/shard_rebalancer-*.sql
 %{pginstdir}/share/extension/shard_rebalancer.control
+%ifarch ppc64 ppc64le
+ %else
+ %if %{pgmajorversion} >= 11 && %{pgmajorversion} < 90
+   %if 0%{?rhel} && 0%{?rhel} <= 6
+   %else
+     %{pginstdir}/lib/bitcode/shard_rebalancer*.bc
+     %{pginstdir}/lib/bitcode/shard_rebalancer/*.bc
```
hanefi, line 56: @velioglu I manually investigated the outputs and added these paths here. If I were to add another line similar to https://github.com/citusdata/packaging/commit/7589e5907ba67de8899e50f33005b203a509a484
```
      %{pginstdir}/lib/bitcode/shard_rebalancer/*/*.bc
```

there does not exist any file matching this wildcards and the build fails.
Locally building and investigating if the rpm package contains all the `*.bc` files is really hard

hanefi:
To test that this packages include all the necessary files I did the following:
- Created a new ol/7 instance. 
- Installed pg with JIT support and tested if it works as intended

```
postgres=# explain select sum(id) from local;
                                     QUERY PLAN
-------------------------------------------------------------------------------------
 Finalize Aggregate  (cost=8706.88..8706.89 rows=1 width=8)
   ->  Gather  (cost=8706.67..8706.88 rows=2 width=8)
         Workers Planned: 2
         ->  Partial Aggregate  (cost=7706.67..7706.68 rows=1 width=8)
               ->  Parallel Seq Scan on local  (cost=0.00..6873.33 rows=333333 width=4)
 JIT:
   Functions: 6
   Options: Inlining false, Optimization false, Expressions true, Deforming true
(8 rows)
```

- Installed citus-enterprise
- Installed shard rebalancer
- Used the multitenant use case steps in our docs
- Observed that `rebalance_table_shards` finishes successfully


```
postgres=# SELECT rebalance_table_shards('companies');
 rebalance_table_shards
------------------------

(1 row)
```

PR #285, opened 2019-02-22
Release Redhat CitusHA 2.0.0

thanodnl:


thanodnl:
There are a lot of compiler warnings, previous release seem to have them as well, therefore I am inclined to move forward

lfittl (approved):
Tested and works correctly, including completing Citus HA quickstart guide with packages on RHEL 7.

lfittl:
@thanodnl Regarding the build failures - seems like we may need to adjust the configuration for Fedora 25/26?

thanodnl:
The images for pg11 seem to be missing for `fedora/25` and `fedora/26`, seeing to rebuild them it seems that the rpm's for pg11 are not available for both, possibly because they are both end of life.

Does it still make sense to build packages for os's that are EOL?

lfittl:
@thanodnl Good question - I don't think we have enough data yet to say which OS would be needed vs not needed. My gut feeling would be to just do what unblocks the build for now (and maybe also make sure to build for newer releases, e.g. I noticed `ubuntu/bionic` wasn't in the meta GitHub issue you have)

cc @jasonmp85 for any additional notes on which releases we should support for packaging.

jasonmp85:
The `os-list.csv` file is pretty much accurate, except for the aforementioned Fedora thing:

https://github.com/citusdata/packaging/blob/4b45746ae4c521383586e04309fd7f9b9a4d2d4e/os-list.csv

We dropped Fedora with PostgreSQL 11's release, as (a) it adds and drops a release every quarter, which is burdensome, (b) we're not sure how much use there is so we decided to drop and see if anyone yelps (they haven't), and (c) maintaining it after the JIT changes became quite difficult (LLVM/Clang backports/versions, etc. got to be too leaky an abstraction for me to care).

If it's got some sort of high-profile user or need, then sure, we can put some time into it. Last I looked that wasn't the case, though.

thanodnl:
I have a meeting with @velioglu later today to push this forward.

thanodnl:
As discussed with @velioglu we will remove the fedora packages from the build matrix until we find a customer that requires fedora images. As per conversation, we don't have citus packages either.

PR #306, opened 2019-04-18
Redhat ha 2.1.0

thanodnl:
Builds require tag on repo, will restart after changelog commit is tagged and released.

redhat 2.0.1 commit is in there, never got released. Will be superseded by this release, but change in that commit is still needed on master, hence using that as a base.

### DIFF
--- a/citus-ha.spec
+++ b/citus-ha.spec
@@ -8,11 +8,11 @@ Summary:	Auto-HA support for Citus
 Name:		%{sname}%{?pkginfix}_%{pgmajorversion}
 Provides:	%{sname}_%{pgmajorversion}
 Conflicts:	%{sname}_%{pgmajorversion}
-Version:	2.0.1
+Version:	2.1.0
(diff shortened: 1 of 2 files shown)


citus-ha.spec
```diff
@@ -8,11 +8,11 @@ Summary:	Auto-HA support for Citus
 Name:		%{sname}%{?pkginfix}_%{pgmajorversion}
 Provides:	%{sname}_%{pgmajorversion}
 Conflicts:	%{sname}_%{pgmajorversion}
-Version:	2.0.0
+Version:	2.1.0
 Release:	1%{dist}
 License:	AGPLv3
```
lfittl: Nitpick, but this is not correct (not that it matters for now, but this should read `Proprietary`)

lfittl (approved):

citus-ha.spec
thanodnl: good catch will change

PR #309, opened 2019-05-01
Redhat pgautofailover 1.0.0

thanodnl:
I have problems running locally from a branch
```
error: %changelog not in descending chronological order
```
which is fascinating as there is only 1 entry.


pg_auto_failover.spec
```diff
@@ -0,0 +1,66 @@
...
+%global sname pg_auto_failover
+%global extname pgautofailover
+
+Summary:	todo
+Name:		%{sname}%{?pkginfix}_%{pgmajorversion}
+Provides:	%{sname}_%{pgmajorversion}
+Conflicts:	%{sname}_%{pgmajorversion}
+Version:	1.0.0
+Release:	1%{dist}
+License:	todo
```
thanodnl: @lfittl can you add the license here?

pg_auto_failover.spec
```diff
@@ -0,0 +1,66 @@
+%global pgmajorversion 11
+%global pgpackageversion 11
+%global pginstdir /usr/pgsql-%{pgpackageversion}
+%global sname pg_auto_failover
+%global extname pgautofailover
+
+Summary:	todo
```
thanodnl: needs to be filled out once we have our 1 liner

pkgvars
```diff
@@ -0,0 +1,6 @@
+pkgname=pg_auto_failover
+hubproj=citus-ha
+pkgdesc=''
```
thanodnl: to be filled out with the oneliner

pkgvars
```diff
@@ -0,0 +1,6 @@
+pkgname=pg_auto_failover
+hubproj=citus-ha
```
thanodnl: needs to change once we have our repo

pg_auto_failover.spec
```diff
@@ -0,0 +1,66 @@
+%global pgmajorversion 11
+%global pgpackageversion 11
+%global pginstdir /usr/pgsql-%{pgpackageversion}
+%global sname pg_auto_failover
```
lfittl: In the blog post I had used `pg-auto-failover-10_11` as the package name to be installed. To keep it simple, could we use "pg-auto-failover" for package names? (that may work for both RPM and DEB)

pg_auto_failover.spec
lfittl: Postgres extension for automated failover and high-availability

pg_auto_failover.spec
lfittl: PostgreSQL License

pg_auto_failover.spec
thanodnl: I used the `_` here to stay consistent with [pg_cron](https://github.com/citusdata/packaging/blob/redhat-cron/pg_cron.spec#L4) but I am ok with this as well.

your blogpost needs to change a bit I think as the package I get is named like this: `pg-auto-failover_10-1.1.0.citus-0.0.pre.20190502.76c4628.el7.x86_64.rpm`
Don't pay attention to `1.1.0` that will become `1.0.0` when we release. but the pg version is separated with a `_`, guess due to line 9-10.

good thing is that the package builds locally after this change :)

thanodnl: guess your name is actually correct, citus-ha packages were having the version in the releases as well, I based this release script on the ones we used there. albeit, without the `-` before the version.

.travis.yml
```diff
@@ -0,0 +1,35 @@
+sudo: required
+services: [ docker ]
+language: perl
+perl: "5.18"
+env:
+  global:
+    - PKG_REPOTYPE=enterprise
```
DimCitus: Do we target community? Maybe later on Monday?

DimCitus (approved):

.travis.yml
thanodnl: good catch, that is where the packages will by pushed, should be community

PR #310, opened 2019-05-06
[Debian] Release pg_auto_failover 1.0.1

thanodnl:
Smoke test passed


debian/changelog
```diff
@@ -1,3 +1,13 @@
+pg-auto-failover (1.0.1) testing; urgency=low
```
DimCitus, line 1: The `debian/changelog` should be about the debian packaging only, so all we need here is a line with

  * New upstream release.

DimCitus (changes requested):

DimCitus (approved):

PR #311, opened 2019-05-06
[Redhat] Release pg_auto_failover 1.0.1

thanodnl:
Smoketests passed


pg-auto-failover.spec
```diff
@@ -60,5 +60,10 @@ PATH=%{pginstdir}/bin:$PATH
 
 
 %changelog
+* Mon May 3 2019 - Nils Dijk <nils@citusdata.com> 1.0.1
```
DimCitus, line 63: I think that one should also be a single entry such as:
  - Official release for 1.0.1

DimCitus (changes requested):

DimCitus (approved):

PR #312, opened 2019-05-23
[Debian] Release pg_auto_failover-enterprise 1.0.2

thanodnl:
First release of enterprise.
Source branch is debian-ha, but moved around as we change the release artifact.

### DIFF
--- a/pkgvars
+++ b/pkgvars
@@ -1,6 +1,6 @@
-pkgname=citus-ha
+pkgname=auto-failover-enterprise
 hubproj=citus-ha
-pkgdesc='Citus HA'
-pkglatest=2.1.0
+pkgdesc='Postgres extension for automated failover and high-availability'
+pkglatest=1.0.2
 releasepg=10,11
 versioning=fancy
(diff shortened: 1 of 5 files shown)


pkgvars
```diff
@@ -1,6 +1,6 @@
-pkgname=citus-ha
+pkgname=auto-failover-enterprise
```
DimCitus, line 1: Is `pkgname` supposed to be `pg-auto-failover-enterprise` or `auto-failover-enterprise` as it is now?

DimCitus (approved):
I think there's a typo to solve.

pkgvars
thanodnl: this seems indeed the case, however, the only place where this is used is a regex that changes the `postgres-XX-auto-failover-enterpise` package name to extend it with the major+minor version of the package. Since postgres is in the name there we have decided to not put `pg_` in the extension package.

DimCitus: I see, thanks for explaining!

PR #336, opened 2019-07-31
Release pg_auto_failover-enterprise 1.0.3

mtuncer:


thanodnl (commented):
Looks good to me, lets make sure the packages can be installed as well. Don't know if redhat had the same bug as debian where the enterprise package depends on the oss cli which is in a different repo.

Will approve after you or me verified it will install correctly (we can test this probably most easiest by installing the previous version on a redhat based system).

mtuncer:
I compared the behavior with 1.0.2 
it is the same in both 1.0.2 and 1.0.3

- rpm -i fails to install the package complaining about `postgresql11-server` and `libpq.so`
- we do `curl https://install.citusdata.com/community/rpm.sh | bash`
- rpm -i still fails
- we do `yum install postgresql11-server`
- rpm -i succeeds

thanodnl:
could you also try the enterprise repository? That was the one having issues on debian. If that installation (including the cli) works we can land

thanodnl:
Just verified there is no problem here with the enterprise repo, lets ship it.

PR #337, opened 2019-08-01
Release pg_auto_failover-enterprise 1.0.3

mtuncer:


thanodnl:
Last time I tried installing the debian packages for enterprise there was a problem with the dependencies. I don't see any changes for that now. Did you test installing the packages after building them locally, and does that work?

mtuncer:
No I did not test debians
Let me do that

debian/changelog
```diff
@@ -1,3 +1,24 @@
+pg-auto-failover-enterprise (1.0.3) testing; urgency=low
+
+  * Fix Terminate conflicting queries at failover time (#194)
```
thanodnl: debian changelogs should not contain the changelog of the application, but instead the changelog to the packaging. I have made that mistake in the past as well :).

Hence the boring changlog here that reads something like `Official release of <package> <version>`

thanodnl (changes requested):

mtuncer:
well I was not able to install pg_auto_failover package via dpkg -i. It complained about missing dependencies.
I run `apt install --fix-broken` after the failure at it installed fine.

thanodnl:
Ok, let me check and reproduce this on debian

thanodnl (approved):

PR #348, opened 2019-09-24
Redhat pgautofailover 1.0.5

mtuncer:


pg-auto-failover.spec
```diff
@@ -8,11 +8,11 @@ Summary:	Postgres extension for automated failover and high-availability
...
-Version:	1.0.4
+Version:	1.0.5
 Release:	1%{dist}
 License:	PostgreSQL
 Group:		Applications/Databases
-Source0:	https://github.com/citusdata/pg_auto_failover/archive/v1.0.4.tar.gz
+Source0:	https://github.com/citusdata/pg_auto_failover/archive/v1.0.5.tar.gz
 URL:		https://github.com/citusdata/pg_auto_failover
 BuildRequires:	postgresql%{pgmajorversion}-devel postgresql%{pgmajorversion}-server libxml2-devel
 BuildRequires:	libxslt-devel openssl-devel pam-devel readline-devel
```
DimCitus, line 18: I don't see python-sphinx in the BuildRequires, is that expected?

DimCitus (approved):

pg-auto-failover.spec
mtuncer: it also requires a recent (1.8.+ I believe) version that can only be installed via pip. At the moment we rely on our own images. We should be fine IMO.

PR #354, opened 2019-10-15
Bump citus debian Version 9.0.0

hanefi:
- Introduced debian/buster.
- Removed ubuntu/trusty as there is no PGDG PostgreSQL 12 repository for it
- Bumped citus tools to use the new docker hub repo for packaging images


jasonmp85 (changes requested):
This looks good _except_ that I'm guessing the reason the `tools` version is bumped is that no one knew how to use the `citusdata` org while I was out (though I swear I got pinged and gave permissions to someone…)

If that's the case, I'd like us to continue using `citusdata` (just for continuity reasons) and I can fix the permissions issues tomorrow, rather than forking our Docker org.

Beyond that, this looks good and :shipit:

hanefi:
> the reason the tools version is bumped is that no one knew how to use the citusdata org while I was out

There are several people with push access now, and sadly I am not one of them. I needed to build and push the packaging images, and I was unable to do that. That is one of the reasons why we created the new `citus` org.

I'd also like to keep using `citusdata` org. I think it should not block us any longer and we should ship the packages now. We can easily bump tools version in the next release to change the docker org later.

PR #356, opened 2019-10-18
Bump Debian Enterprise Version 9.0.0 

hanefi:
- Introduced debian/buster.
- Removed ubuntu/trusty as there is no PGDG PostgreSQL 12 repository for it
- Bumped citus tools to use the new docker hub repo for packaging images


jasonmp85 (changes requested):
Same feedback as https://github.com/citusdata/packaging/pull/354#pullrequestreview-307681415

So basically a :shipit: once the Docker images are back in the right org and tools is updated to reflect that.

hanefi:
Same answer to your feedback as https://github.com/citusdata/packaging/pull/354#issuecomment-546877523

I plan to bump the tools version in the next release. Before then, I think we should not be blocked any more on the docker org changes

PR #358, opened 2019-10-21
Update deb and rpm scripts to use PG12

hanefi:
Update all the scripts to use PG12 repositories instead of 11.

We may need to do some testing in some repositories before landing the changes here.

- [x] ~We have not yet released all packages, so this PR should be blocked until we are able to test it properly~
	- The packaging of `9.0.1` is complete. I verified that the scripts in this branch works in all the new distros we started to support.
- [x] OL/EL 8 PGDG repositories have issues, ~we should wait until it gets fixed.~
	- The script works fine in the OL/EL 8. Once we start creating packages for these distros the changes here will hopefully work.
- [ ] CentOS 8 advises us to use `dnf` instead of `yum` to manage packages. We may consider doing that in these scripts as well
	- I am discarding this idea for now, as this means we will diverge from the package cloud scripts and we need to spend some more time on this to get it working.

### DIFF
--- a/community-nightlies/deb.sh
+++ b/community-nightlies/deb.sh
@@ -163,6 +163,9 @@ detect_codename ()
       9)
         codename='stretch'
         ;;
+      10)
+        codename='buster'
+        ;;
       wheezy)
         codename="${dist}"
         ;;
@@ -170,6 +173,9 @@ detect_codename ()
         codename="${dist}"
         ;;
       stretch)
+        codename="${dist}"
+        ;;
+      buster)
         codename="${dist}"
         ;;
       *)
--- a/community-nightlies/deb.sh
+++ b/community-nightlies/deb.sh
@@ -56,9 +56,9 @@ curl_check ()
 
 pgdg_check ()
 {
-  echo "Checking for postgresql-11..."
-  if apt-cache show postgresql-11 &> /dev/null; then
-    echo "Detected postgresql-11..."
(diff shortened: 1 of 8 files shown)


jasonmp85:
Noticed this draft PR and figured I'd point you to the upstream packagecloud scripts for reference…

  * https://packagecloud.io/citusdata/community/install
  * https://packagecloud.io/citusdata/enterprise/install

You'll probably need to be logged in to a packagecloud account marked as a collaborator for the second link, but both scripts are somewhat similar. We based our original scripts on those made by packagecloud (I asked for a GitHub repo/license a long time ago to no avail), so I will sometimes check upstream to see whether they've made any changes or improvements we want to port over.

The PGDG stuff in our script is obviously something we added for our PostgreSQL use-case.

And some other coworkers seemed unclear on which branch actually controls these scripts so I figured I'd spell it out that `gh-pages` is indeed where the scripts come from. If we merge to that branch, the changes go live _immediately_ so testing is important.

There are Amazon and Cloudflare intermediary services (cloudfront, etc.) that sit between GitHub and the end user, mostly for stuff like SSL and our domain, etc., but the scripts are ultimately just a static page/file served by GitHub Pages.

hanefi:
I went over the packagecloud scripts in the links @jasonmp85 shared. There does not seem to be any change in neither rpm or deb scripts.

velioglu (approved):

PR #360, opened 2019-10-30
Bump redhat Citus 9.0.1

hanefi:


citus.spec
```diff
@@ -1,17 +1,17 @@
-%global pgmajorversion 10
-%global pgpackageversion 10
+%global pgmajorversion 11
```
velioglu, line 1: Do we still need pgmajorversion check at the line 66 ?

velioglu (approved):

citus.spec
hanefi: I guess we don't. I'll remove that unnecessary check before merging.

velioglu (approved):

PR #363, opened 2019-10-30
Bump redhat Citus Enterprise 9.0.1

hanefi:


citus-enterprise.spec
```diff
@@ -1,5 +1,5 @@
-%global pgmajorversion 10
-%global pgpackageversion 10
+%global pgmajorversion 11
```
velioglu, line 1: Do we still need pgmajorversion check at the line 67 ?

velioglu (approved):

citus-enterprise.spec
hanefi: I guess we don't. I'll remove that unnecessary check before merging.

velioglu (approved):

PR #372, opened 2019-11-28
Bump citus debian Version

onurctirtir:


velioglu (approved):
Discussing with @metdos @onderkalaci  and @onurctirtir we decided that releasing packages with two warnings on the travis is ok. One of them is related to https://github.com/citusdata/citus/pull/2844/commits/e5cd298a987fd4b1b0804e61de21863031bf2078 and the other one is observed by engine developers in their local env as well. So, not related to packaging directly.

PR #373, opened 2019-12-09
Bump Citus to 9.1.0

onurctirtir:


META.json
```diff
@@ -2,15 +2,15 @@
...
    "description": "Citus horizontally scales PostgreSQL across commodity servers using sharding and replication. Its query engine parallelizes incoming SQL queries across these servers to enable real-time responses on large datasets.",
-   "version": "9.0.1",
+   "version": "9.1.0",
    "maintainer": "\"Citus Data\" <packaging@citusdata.com>",
    "license": "agpl_3",
    "provides": {
       "citus": {
          "abstract": "Citus Distributed Database",
-         "file": "src/backend/distributed/citus--9.0-1.sql",
+         "file": "src/backend/distributed/citus--9.1-0.sql",
```
hanefi: We no longer provide this migration script in this directory. Should we change it into a valid file, or change it into something that does not change over version updates maybe?

hanefi: It can be worth checking PGXN specs at https://pgxn.org/spec/#provides :

> file:
The value must contain a relative file path from the root of the distribution to the file containing the extension. The path must be specified with unix conventions. Required.

hanefi (changes requested):
One minor comment. Please update the `file` value in the spec.

META.json
onurctirtir: done in 505e630cf071e3c43db6366fd4655431aa16dcbf

hanefi (approved):
:shipit:

PR #374, opened 2019-12-11
Fix pgdg paths and use dnf if needed

onurctirtir:
This pr aims to:
* Fix pgdg paths in rpm.sh scripts
* Add dnf call for RedHat/CentOS 8 to disable the built-in PostgreSQL module as recommended

### DIFF
--- a/community/rpm.sh
+++ b/community/rpm.sh
@@ -38,6 +38,10 @@ pgdg_check ()
   else
     echo -n "Installing pgdg12 repo... "
 
+    if [ "${dist}" = "8" ]; then
+      dnf -qy module disable postgresql
+    fi
+
     yum install -d0 -e0 -y "${repo_url}"
     echo "done."
   fi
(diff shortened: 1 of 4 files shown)


community/rpm.sh
```diff
@@ -153,7 +157,7 @@ detect_repo_url ()
 {
   # set common defaults used by most flavors
   family='redhat'
-  family_short='rhel'
+  family_short='EL'
```
hanefi, line 160: Can we lowercase this?

enterprise/rpm.sh
```diff
@@ -183,7 +187,7 @@ detect_repo_url ()
 {
   # set common defaults used by most flavors
   family='redhat'
-  family_short='rhel'
+  family_short='EL'
```
hanefi, line 190: `s/EL/el/` ?

hanefi (changes requested):
One minor comment, otherwise :shipit: ?

hanefi (approved):
Seeing that the URL is case sensitive, I withdraw my previous comment to lowercase family names.

🚀

PR #378, opened 2019-12-19
Bump Citus to 9.1.1

onurctirtir:

PR #387, opened 2020-02-10
Bump Citus to 9.2.0

onurctirtir:


META.json
```diff
@@ -2,15 +2,15 @@
...
    "description": "Citus horizontally scales PostgreSQL across commodity servers using sharding and replication. Its query engine parallelizes incoming SQL queries across these servers to enable real-time responses on large datasets.",
-   "version": "9.1.1",
+   "version": "9.2.0",
    "maintainer": "\"Citus Data\" <packaging@citusdata.com>",
    "license": "agpl_3",
    "provides": {
       "citus": {
          "abstract": "Citus Distributed Database",
-         "file": "src/backend/distributed/sql/citus--9.0-2--9.1-1.sql",
+         "file": "src/backend/distributed/sql/citus--9.0-2--9.2-0.sql",
```
JelteF: I don't think the file mentioned here actually exists. Maybe something like this?
```suggestion
         "file": "src/backend/distributed/sql/citus--9.1-1--9.2-0.sql",
```

onurctirtir: Could be, but actually that "file" field seems (btw, not exactly sure) redundant as @hanefi mentioned in previous release https://github.com/citusdata/packaging/pull/378/files#r359812702. Thinking about it, we already execute each migration script sequentially. But applying the changes you proposed seems more convenient.

onurctirtir: applied your changes

PR #397, opened 2020-02-14
Bump Citus to 9.2.1

onurctirtir:


META.json
```diff
@@ -2,15 +2,15 @@
...
    "description": "Citus horizontally scales PostgreSQL across commodity servers using sharding and replication. Its query engine parallelizes incoming SQL queries across these servers to enable real-time responses on large datasets.",
-   "version": "9.2.0",
+   "version": "9.2.1",
    "maintainer": "\"Citus Data\" <packaging@citusdata.com>",
    "license": "agpl_3",
    "provides": {
       "citus": {
          "abstract": "Citus Distributed Database",
-         "file": "src/backend/distributed/sql/citus--9.1-1--9.2-0.sql",
+         "file": "src/backend/distributed/sql/citus--9.1-1--9.2-1.sql",
```
hanefi: AFAIK, there is no such file. Did you mean `citus--9.2-0--9.2-1.sql` or maybe `citus--9.2-1--9.2-2.sql` ?

hanefi (changes requested):
one minor issue with `file` field in `META.json`

hanefi (approved):
After more investigation, the `file` field used to be wrong, and it is now fixed.

so :shipit:

PR #400, opened 2020-02-17
Fix: re-add file field to META.json

onurctirtir:
As we could not upload pgxn package in prev pr, I am re-adding that dummy file field


hanefi (approved):
:shipit:

PR #415, opened 2020-03-18
Update images to support new builds

JelteF:
This adds two things:
1. It adds cmake (and which) to the build to support building the latest
   version of Citus.
2. It forwards the PACKAGE_ENCRYPTION_KEY environment variable to debbuild.
   This is needed for the new encrypted enterprise packages.


hanefi (approved):
I created packages for several (if not all) distros and they seem to be able to create encrypted packages.

:shipit:

PR #416, opened 2020-03-18
Changes for debian packages to be hosted on microsoft infrastructure

JelteF:
Things this does:
1. Encrypt the `.so`, `.bc` and `.control` files, so the package is effectively
   unusable without decrypting them
2. Adds a `citus-enterprise-pg-$PGVERSION-setup` script that needs to be run
   after install for the installation to complete
3. Add a postinst script, telling the user to run the required script
4. Add a prerm script to cleanup the files that the
5. Disable libcurl integration, so statistics collection is compiled out

The `citus-enterprise-pg-$PGVERSION-setup` does three things:
1. Require the user to accept software terms
2. Require the user to acknowledge the risks of unencrypted traffic
3. Require the user to enter the licence key they received during purchase of
   the product

### DIFF
--- a/debian/citus-setup
+++ b/debian/citus-setup
@@ -93,7 +93,6 @@ done
 if ! gpg --output "/dev/null" \
         --batch --no-tty --yes --quiet \
         --passphrase "$CITUS_LICENSE_KEY" \
-        --pinentry loopback \
         --homedir "$temp_gnupghome" \
         --decrypt "$(head -n 1 "$secret_files_list").gpg" 2> /dev/null; then
     echo "ERROR: Invalid license key supplied"
@@ -110,7 +109,6 @@ while read -r path_unencrypted; do
     gpg --output "$path_unencrypted" \
         --batch --no-tty --yes --quiet \
         --passphrase "$CITUS_LICENSE_KEY" \
-        --pinentry loopback \
         --homedir "$temp_gnupghome" \
         --decrypt "$path_encrypted"
 
--- a/debian/citus-setup
+++ b/debian/citus-setup
@@ -1,0 +1,123 @@
+#!/bin/sh
+
+set -eu
+
+pg_version=
+libdir="/usr/lib/postgresql/$pg_version/lib"
+secret_files_list="$libdir/citus_secret_files.metadata"
+
+# Make sure the script is being run as root
+if [ "$(id -u)" -ne "0" ]; then
+    echo "ERROR: citus-enterprise-pg-$pg_version-setup needs to be run as root"
+    echo "HINT: try running \"sudo citus-enterprise-pg-$pg_version-setup\" instead"
+    exit 1
+fi
+
+
+eula_text="
+Your use of this software is subject to the terms and conditions of the license
+agreement by which you acquired this software. If you are a volume license
+customer, use of this software is subject to your volume license agreement.
+You may not use this software if you have not validly acquired a license for
+the software from Microsoft or its licensed distributors.
+
+Do you accept these terms? y/N"
+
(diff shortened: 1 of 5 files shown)


hanefi:
If we want to keep using Travis we need to bump tools, and update the `.travis.yml` accordingly

https://github.com/citusdata/packaging/blob/27f2266479c8d4664b0f99c466ff894ce1238f41/.travis.yml#L21

In case this will only be used in DevOps, we still need to create a tag and make sure we use the tag there as well

JelteF:
Good catch, I'll fix that

debian/citus-setup
```diff
@@ -0,0 +1,121 @@
...
+2. The ssl or ssl_ciphers Postgres configuration option is set to something
+   other than the default.
+
+In these cases it is assumed the operator has set up appropriate security
+themselves.
+
+So, with the default settings Citus clusters are not safe from
+Man-In-The-Middle attacks. To secure the traffic completely you need to follow
+the practices outlined here:
+https://citus.com/some-page-on-secure-setup
```
hanefi, line 64: We need to update the placeholder here before merge

debian/encrypt_files.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+set -euo pipefail
+
+PACKAGE_ENCRYPTION_KEY="${PACKAGE_ENCRYPTION_KEY:-}"
+if [ -z "$PACKAGE_ENCRYPTION_KEY" ]; then
+    echo "ERROR: The PACKAGE_ENCRYPTION_KEY environment variable needs to be set"
+    echo "HINT: If trying to build packages locally, just set it to 'abc' or something"
```
hanefi, line 7: thanks for the hint, should we remove this before merge?

debian/citus-setup
```diff
@@ -0,0 +1,121 @@
...
+CITUS_ACCEPT_ENCRYPTION_WARNING="${CITUS_ACCEPT_ENCRYPTION_WARNING:-}"
+while [ -z "$CITUS_ACCEPT_ENCRYPTION_WARNING" ]; do
+    echo "$ssl_warning_text"
+    read -r CITUS_ACCEPT_ENCRYPTION_WARNING
+done
+
+case "$CITUS_ACCEPT_ENCRYPTION_WARNING" in
+    y|Y|Yes|YES|yes );;
+    * )
+        echo "ERROR: Warning about encrypted traffic must be accepted before installing"
```
hanefi, line 79: I think this is more of a disclaimer, and I think `Warning` may sound too harsh here.

hanefi (changes requested):
It works perfectly fine, need to polish some user-facing text before merge.

some minor issues:
- Placeholder url
- Outdated `9.2.0` version in `pkgvars`
- One hint that better be removed :)
- I suggest s/warning/disclaimer/ in one place

I tested on debian buster, could not access anything before running `citus-enterprise-pg-$pg_version-setup`, and it decrypted once I supplied a correct key

hanefi:
Suggestion: let's add a dummy `PACKAGE_ENCRYPTION_KEY` env in `.travis.yml` so that we can test changes in CI before making changes in this branch.

We usually follow steps in [Circle CI docs](https://docs.travis-ci.com/user/environment-variables/) to encrypt environment variables, but there is no need here I guess

debian/encrypt_files.sh
JelteF: I think it's useful to keep it in. This is only displayed using package building.

debian/citus-setup
JelteF: Good call, totally forgot about that

JelteF:
Closing this, since we only have a single branch for microsoft packages.

PR #417, opened 2020-03-18
Add Microsoft package requirements for redhat

JelteF:
Same as https://github.com/citusdata/packaging/pull/416, but then for Redhat. Basically the things that are separate files in that PR are all added to the `citus-enterprise.spec` file here. So it's easier to review the other PR.

Things this does:
1. Encrypt the `.so`, `.bc` and `.control` files, so the package is effectively
   unusable without decrypting them
2. Adds a `citus-enterprise-pg-$PGVERSION-setup` script that needs to be run
   after install for the installation to complete
3. Add a postinst script, telling the user to run the required script
4. Add a prerm script to cleanup the files that the
5. Disable libcurl integration, so statistics collection is compiled out

The `citus-enterprise-pg-$PGVERSION-setup` does three things:
1. Require the user to accept software terms
2. Require the user to acknowledge the risks of unencrypted traffic
3. Require the user to enter the licence key they received during purchase of
   the product

NOTE: This also disables debug packages. These can be useful when receiving
coredumps from customers. So at a later point it would be good to figure out a
way to enable them again.

### DIFF
--- a/citus-enterprise.spec
+++ b/citus-enterprise.spec
@@ -3,6 +3,7 @@
 %global pginstdir /usr/pgsql-%{pgpackageversion}
 %global sname citus-enterprise
 %global pname citus
+%global debug_package %{nil}
 
 Summary:	PostgreSQL-based distributed RDBMS
 Name:		%{sname}%{?pkginfix}_%{pgmajorversion}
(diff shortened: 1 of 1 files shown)


citus-enterprise.spec
```diff
@@ -3,6 +3,7 @@
 %global pginstdir /usr/pgsql-%{pgpackageversion}
 %global sname citus-enterprise
 %global pname citus
+%global debug_package %{nil}
```
hanefi, line 6: Understandable change.

Do we intentionally disable debuginfo during package creation here?

If a user runs into any issue, will it be possible for us to reach out to them and show a way to install debug packages later to help us resolve their issues?

citus-enterprise.spec
```diff
@@ -44,6 +45,201 @@ make %{?_smp_mflags}
...
+fi
+
+echo "License key is valid"
+echo "Installing..."
+
+# Decrypt all the encrypted files
+while read -r path_unencrypted; do
+    path_encrypted="\$path_unencrypted.gpg"
+    # decrypt the encrypted file
+    gpg --output "\$path_unencrypted" \
```
hanefi, line 200: My screen is filled with warnings, and I guess theis line is the culprit here.

```
Please enter license key:
verysecure
License key is valid
Installing...
gpg: WARNING: message was not integrity protected
```
The last line was repeated 161 times

citus-enterprise.spec
```diff
@@ -44,6 +45,201 @@ make %{?_smp_mflags}
...
+2. The ssl or ssl_ciphers Postgres configuration option is set to something
+   other than the default.
+
+In these cases it is assumed the operator has set up appropriate security
+themselves.
+
+So, with the default settings Citus clusters are not safe from
+Man-In-The-Middle attacks. To secure the traffic completely you need to follow
+the practices outlined here:
+https://citus.com/some-page-on-secure-setup
```
hanefi, line 155: Should we update the placeholder URL before merge?

citus-enterprise.spec
```diff
@@ -44,6 +45,201 @@ make %{?_smp_mflags}
...
+CITUS_ACCEPT_ENCRYPTION_WARNING="\${CITUS_ACCEPT_ENCRYPTION_WARNING:-}"
+while [ -z "\$CITUS_ACCEPT_ENCRYPTION_WARNING" ]; do
+    echo "\$ssl_warning_text"
+    read -r CITUS_ACCEPT_ENCRYPTION_WARNING
+done
+
+case "\$CITUS_ACCEPT_ENCRYPTION_WARNING" in
+    y|Y|Yes|YES|yes );;
+    * )
+        echo "ERROR: Warning about encrypted traffic must be accepted before installing"
```
hanefi, line 170: Similar to https://github.com/citusdata/packaging/pull/416/files#r395156036 :
> I think this is more of a disclaimer, and I think Warning may sound too harsh here.

citus-enterprise.spec
```diff
@@ -44,6 +45,201 @@ make %{?_smp_mflags}
...
+
+%preun
+libdir="%{pginstdir}/lib"
+
+secret_files_list="$libdir/citus_secret_files.metadata"
+
+# Cleanup all de decrypted files since these are not managed by the package
+# manager and would be left around otherwise
+while read -r path_unencrypted; do
+    rm -f "$path_unencrypted"
```
hanefi, line 241: Info: the `rm` here causes a warning during package builds, but that is not really harmful in any way
> W: dangerous-command-in-%preun rm

hanefi (approved):
Looks good and creates packages perfectly.

Several questions, and remarks that are not really blockers.

citus-enterprise.spec
JelteF: It turns out that the debug symbols should actually be in the main package now. The reason that I had to add this line was because `rpmbuild` could not exctract the debug symbols anymore, since the .so file is now encrypted. This means that they are also not stripped out of the `.so` file, so they should be in the `.so` file after decrypting.

citus-enterprise.spec
JelteF: Good catch, which OS was this?
According to the answer here, it's fied by explicitely setting AES256 as the cipher algo (which seems smart anyway): https://superuser.com/a/633716/242593

I did that here now: https://github.com/citusdata/packaging/pull/418/commits/5d2ca535af5ee71a3b85e8ded3719af9c59d4de3

citus-enterprise.spec
JelteF: Done in https://github.com/citusdata/packaging/pull/418/commits/d9afe5ec54149ead58d18155b5c1cb5c5fded9b7

citus-enterprise.spec
JelteF: Done in https://github.com/citusdata/packaging/pull/418/commits/d9afe5ec54149ead58d18155b5c1cb5c5fded9b7

PR #424, opened 2020-03-26
Use GCC8 on CentOS when available

JelteF:
For security compliance we cannot use a GCC lower than 4.8. The default GCC on
centos6 and oraclelinux6 are GCC 4.4, so these don't comply.
This tries to add the GCC8 to the image using the devtoolset-8-gcc packge. This
is not available everywhere.


hanefi (approved):
LGTM,
lets build and push to docker hub

PR #425, opened 2020-03-26
Check that GCC used to compile has a bigger version than 4.8.0

JelteF:
This is a check to ensure that Citus Enterprise is not built with a GCC version
lower than 4.8.0. This is mostly used as a way to ensure that el/6 and ol/6 are
not accidentily built before #424 is merged.

### DIFF
--- a/citus-enterprise.spec
+++ b/citus-enterprise.spec
@@ -37,7 +37,15 @@ commands.
 %setup -q -n %{sname}-%{version}
 
 %build
-%configure PG_CONFIG=%{pginstdir}/bin/pg_config --with-extra-version="%{?conf_extra_version}"
+
+currentgccver="$(gcc -dumpversion)"
+requiredgccver="4.8.0"
+if [ "$(printf '%s\n' "$requiredgccver" "$currentgccver" | sort -V | tail -n1)" = "$requiredgccver" ]; then
+    echo ERROR: At least GCC version "$requiredgccver" is needed
+    exit 1
+fi
+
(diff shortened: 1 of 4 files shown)


citus-enterprise.spec
```diff
@@ -37,7 +37,15 @@ commands.
...
-%configure PG_CONFIG=%{pginstdir}/bin/pg_config --with-extra-version="%{?conf_extra_version}"
+
+currentgccver="$(gcc -dumpversion)"
+requiredgccver="4.8.0"
+if [ "$(printf '%s\n' "$requiredgccver" "$currentgccver" | sort -V | tail -n1)" = "$requiredgccver" ]; then
+    echo ERROR: At least GCC version "$requiredgccver" is needed
+    exit 1
+fi
+
+%configure PG_CONFIG=%{pginstdir}/bin/pg_config --with-extra-version="%{?conf_extra_version}" CC=$(command -v gcc)
```
hanefi, line 48: I am curious, why do we need `CC=$(command -v gcc)` here?

hanefi (approved):
LGTM, just one question out of curiosity. I am sure it has a valid and logical answer

hanefi (changes requested):
I just realized failed builds on `EL/6`

LGTM after we fix the issues there. Blocked until https://github.com/citusdata/packaging/pull/424 is merged I guess

citus-enterprise.spec
JelteF: Mainly to be sure from the build output that we are using the right one. I don't think it's really needed per se, because `pg_config --cc` outputs `gcc --std=gnu99`. If it would use an absolute path though it would be an issue.

PR #427, opened 2020-03-27
Fix compilation flags for microsoft packages

JelteF:
--with-libcurl was ignored, because pg_extbuilder wants all the configure arguments in a single argument (which is weird)


onurctirtir (approved):
as long as you tested the flag changes against package build, I'm okay with the changes

PR #440, opened 2020-04-01
Preserve UNENCRYPTED_PACKAGE envvar in debuild

JelteF:
Forward ~MICROSOFT_PACKAGE~ UNENCRYPTED_PACKAGE environment variable to debuild. This is needed for
packages to handle Microsoft builds correctly.

PR #441, opened 2020-04-01
Merge all enterprise packaging in a single branch

JelteF:
We now have three branches to release Citus Enterpris, i.e `redhat-enterprise`, `debian-enterprise` and `microsoft-enterprise`. This merges them all in a single branch. It does require #440, otherwise no encrypted packages will ever be built for debian based distros.


.travis.yml
```diff
@@ -25,8 +38,39 @@ env:
     - TARGET_PLATFORM=ubuntu/bionic
     - TARGET_PLATFORM=ubuntu/xenial
 before_install:
-  - export PACKAGE_ENCRYPTION_KEY=verysecret
-  - git clone -b v0.7.14 --depth 1 https://github.com/citusdata/tools.git
+  - git clone -b simplify-branching --depth 1 https://github.com/citusdata/tools.git
```
JelteF: This should be changed once https://github.com/citusdata/tools/pull/62 is merged

.travis.yml
```diff
@@ -25,8 +38,39 @@ env:
...
   - sudo make -C tools install
+  - |
+    if [ -n "${MICROSOFT_PACKAGE:-}" ]; then
+    export PACKAGE_ENCRYPTION_KEY=verysecret
+    else
+    export PACKAGING_SECRET_KEY="$(openssl aes-256-cbc -K $encrypted_a39e3f1ee8ba_key -iv $encrypted_a39e3f1ee8ba_iv -in signing_key.asc.enc -d | base64)"
+    fi
 install: true
-script: "export PACKAGE_ENCRYPTION_KEY=verysecret && citus_package -p ${TARGET_PLATFORM} 'local' 'release'"
+script: |
```
hanefi, line 50: As far as I understand, you wanted to change the CI behaviour here depending on MS and non-MS packages.

MS package: we will try to build the packages and we will never attempt to push them to PackageCloud. It will be sufficient to create the package

Non-MS packages: similar to what we already have, we will create the packages, and upload them to packagecloud once the PR is merged.

Is my understanding correct?

hanefi (approved):
LGTM,

I am in favor of creating another PR for merging all community packages.

I should also note that we should merge #440, create a new tag and update the `simplify-branching` with said tag

.travis.yml
JelteF: Yes that is correct. One thing to note is that teh MS packages also test that the encryption is working.

.travis.yml
onurctirtir: we should not forget to revert this back before merging.

onurctirtir (approved):
My local environment doesn't allow me to build os packages locally hence I couldn't test the pr. But after merging this one and #440, I will test on ci by executing update_os_package script

PR #449, opened 2020-04-10
Update PGDG rpm repo paths

hanefi:
Recently PGDG rpm repository paths changed. I made the necessary changes to the `rpm.sh` files

- Removed fedora from the supported OS list 
- Removed unused variables

To test the script you can use the following to use the scripts in this branch:
```
curl https://raw.githubusercontent.com/citusdata/packaging/update-pgdg-rpm-paths/community/rpm.sh | sudo bash
```

### DIFF
--- a/community-nightlies/rpm.sh
+++ b/community-nightlies/rpm.sh
@@ -155,7 +155,8 @@ finalize_zypper_repo ()
 
 detect_repo_url ()
 {
-  # set common defaults used by most flavors
+  # set common defaults used by all flavors
+  # Various repositories of different flavors are stored in EL (e.g. Centos and OL)
   family='redhat'
   family_short='EL'
   pkg_dist="${dist}"
--- a/community-nightlies/rpm.sh
+++ b/community-nightlies/rpm.sh
@@ -157,25 +157,11 @@ detect_repo_url ()
 {
   # set common defaults used by most flavors
   family='redhat'
-  family_short='rhel'
+  family_short='EL'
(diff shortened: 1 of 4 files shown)


onurctirtir (commented):
EL 6/7/8 tests with centos failed for me, investigating

edit: I realized that I even don't have a packagecloud enterprise token, trying on community now

EL 6-7-8 & OL 6-7 are ok. I tested the scirpt on centos & oraclelinux. OL 8 errors out as expected

onurctirtir:
I wonder why we removed fedora from the scripts ?

edit: discussed in person and @hanefi noticed me that we don't even release packages for fedora for a long time

onurctirtir (dismissed):
LGTM, but please add a comment for why it does work even if we don't change `family_short`

hanefi:
>I wonder why we removed fedora from the scripts ?

[PG Redhat Download page](https://www.postgresql.org/download/linux/redhat/) contains a note saying that not all Fedora packages are available. We ran into some issues in the past and dropped Fedora support starting from Citus 8.1.

```
The PostgreSQL Yum Repository currently supports:

Red Hat Enterprise Linux
CentOS
Scientific Linux
Oracle Linux
Fedora*
*Note: due to the shorter support cycle on Fedora, all supported versions of PostgreSQL are not available on this platform. We do not recommend using Fedora for server deployments.
```

onurctirtir (approved):

PR #454, opened 2020-04-29
Merge redhat and debian build branch for citus OSS

JelteF:
Same as #441 but for opensource

PR #455, opened 2020-04-29
Merge redhat and debian branches of pgautofailover OSS

JelteF:
Same as #441 and #454, but for pgautofailover OSS


onurctirtir (approved):
Thanks for splitting the pr into seperate commits, after merging, we should test the scripts executing `update_os_packages` script

PR #456, opened 2020-04-29
Merge debian and redhat builds for pgautofailover enterprise

JelteF:
Same as #441, #454 and #455 but for pgautofailover enterprise


onurctirtir (approved):
Thanks for splitting the pr into seperate commits, after merging, we should test the scripts executing `update_os_packages` script

PR #457, opened 2020-04-30
Change the way the license text is shown

JelteF:
This way the terms are always shown, even when doing non interactive install.


onurctirtir (approved):
built a new packages and installed it to test

PR #460, opened 2020-05-01
Install sphinx on all distros with python3

JelteF:
This is needed to build manpages for pgautofailover

### DIFF
--- a/scripts/fetch_and_build_deb
+++ b/scripts/fetch_and_build_deb
@@ -172,6 +172,6 @@ pg_buildext updatecontrol
 procs="$(nproc)"
 mjobs="$(expr $procs + 1)"
 
-DEB_BUILD_OPTIONS="parallel=${mjobs}" debuild --preserve-envvar CONF_EXTRA_VERSION --preserve-envvar UNENCRYPTED_PACKAGE --preserve-envvar PACKAGE_ENCRYPTION_KEY -uc -us -B --lintian-opts --profile debian --allow-root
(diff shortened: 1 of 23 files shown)


scripts/fetch_and_build_deb
```diff
@@ -172,6 +172,6 @@ pg_buildext updatecontrol
 procs="$(nproc)"
 mjobs="$(expr $procs + 1)"
 
-DEB_BUILD_OPTIONS="parallel=${mjobs}" debuild --preserve-envvar CONF_EXTRA_VERSION --preserve-envvar UNENCRYPTED_PACKAGE --preserve-envvar PACKAGE_ENCRYPTION_KEY -uc -us -B --lintian-opts --profile debian --allow-root
+DEB_BUILD_OPTIONS="parallel=${mjobs}" debuild --prepend-path /usr/local/bin --preserve-envvar CONF_EXTRA_VERSION --preserve-envvar UNENCRYPTED_PACKAGE --preserve-envvar PACKAGE_ENCRYPTION_KEY -uc -us -B --lintian-opts --profile debian --allow-root
```
hanefi, line 175: Is the sphinx binaries installed in `/usr/local/bin` ? Why did we need this line?

I am asking this mainly for understanding if this is something that I missed.

hanefi (approved):
Looks good overall. One question that is only for information sharing purposes.

scripts/fetch_and_build_deb
JelteF: Yes, `pip install ...` installs binaries in `/usr/local/bin` and `debuild` doesn't automatically have that in PATH, so you have to add it like this.

PR #461, opened 2020-05-01
Build manpages on debian

JelteF:
Depends on #460

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ env:
     - TARGET_PLATFORM=debian/stretch
     - TARGET_PLATFORM=ubuntu/bionic
     - TARGET_PLATFORM=ubuntu/xenial
-    - TARGET_PLATFORM=ubuntu/trusty
 before_install:
   - git clone -b v0.7.15 --depth 1 https://github.com/citusdata/tools.git
   - sudo make -C tools install
--- a/debian/rules
+++ b/debian/rules
@@ -10,6 +10,7 @@ clean: debian/control
 override_dh_auto_build:
 	+make -C $(shell pwd)/src/bin/pg_autoctl
 	+pg_buildext build $(shell pwd)/src/monitor build-%v
+	+make man
 
 override_dh_auto_clean:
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -20,7 +20,6 @@ env:
     - TARGET_PLATFORM=debian/stretch
     - TARGET_PLATFORM=ubuntu/bionic
     - TARGET_PLATFORM=ubuntu/xenial
-    - TARGET_PLATFORM=ubuntu/trusty
```
hanefi, line 23: `ubuntu/trusty` EOL'd on April 30, 2019.
Good riddance

hanefi (approved):
LGTM

.travis.yml
```diff
@@ -20,7 +20,6 @@ env:
     - TARGET_PLATFORM=debian/stretch
     - TARGET_PLATFORM=ubuntu/bionic
```
hanefi, line 21: @JelteF can you go over the list of distros here, and add some missing ones if there are any?

**el/6** may have some outdated tooling and create issues
**ol/6** is a bit old, and I do not know why we do not have it here
**ol/8** it is unclear if PackageCloud supports this distro yet.
**Ubuntu/focal** is relatively new and I guess should be added here.

hanefi (approved):
Can you review the list of distros first, and add missing ones?

.travis.yml
JelteF: Tried with the these distros, they all fail: https://travis-ci.org/github/citusdata/packaging/builds/683750588

1. el/6 and ol/6 fail because of a static assert that is unsupported by their ancient gcc. This should be fixed for el/6 once we build new pg10 images, since it will use a new GCC then.
2. ol/8 and ubuntu/focal fail because of missing images on docker

hanefi: > ol/8 and ubuntu/focal fail because of missing images on docker.

I forgot to push the images, sorry. Do you think we should create the packages for those in a separate PR?

Thanks for spending the time to make sure we do not miss anything.

PR #463, opened 2020-05-04
Support encrypting of pgautofailover enterprise

JelteF:


.travis.yml
```diff
@@ -12,23 +12,47 @@ env:
...
+    # Packages on packages.microsoft.com
     - TARGET_PLATFORM=el/6
-    - TARGET_PLATFORM=ol/7
-    - TARGET_PLATFORM=ol/6
+    # Removed for now because it has no gcc4.8+ (which is required for security
+    # compliance)
+    # - TARGET_PLATFORM=ol/6
+    - TARGET_PLATFORM=el/7
+    # Removed for now since we don't have oracle linux repos on microsoft
+    # - TARGET_PLATFORM=ol/7
```
hanefi, line 22: I really like the comments that state the reason why you intentionally left out some distros

.travis.yml
```diff
@@ -12,23 +12,47 @@ env:
...
+    - TARGET_PLATFORM=el/8
     - TARGET_PLATFORM=debian/buster
-    - TARGET_PLATFORM=debian/jessie
     - TARGET_PLATFORM=debian/stretch
+    - TARGET_PLATFORM=debian/jessie
     - TARGET_PLATFORM=ubuntu/bionic
     - TARGET_PLATFORM=ubuntu/xenial
```
hanefi, line 28: `ubuntu/focal` and `ol/8` are missing on both packagecloud and microsoft targets. Is that intentional?

I guess we'll not have the `ol/8` on MS repositories, however for consistency purposes we can also add this to the comments there maybe.

debian/decrypt-files.sh
```diff
@@ -0,0 +1,147 @@
...
+interactive_license=false
+while [ -z "$CITUS_ACCEPT_LICENSE" ]; do
+    interactive_license=true
+    echo "Do you accept these terms? YES/NO"
+    read -r CITUS_ACCEPT_LICENSE
+done
+
+case "$CITUS_ACCEPT_LICENSE" in
+    YES );;
+    y|Y|Yes|yes )
```
hanefi, line 38: This seems really unintuitive to me. I am used to typing a single letter (y or n) when such a prompt appears without actually reading all my options.

If there is a solid reason thy only `YES` is allowed, I'm totally ok with this

JelteF: Legel team prefered it that way. (It's the same for the citus enterprise microsoft packages.)

.travis.yml
JelteF: Resolved in https://github.com/citusdata/packaging/pull/473

.travis.yml
JelteF: Resolved in https://github.com/citusdata/packaging/pull/473

JelteF:
Closing this in favor of Resolved in https://github.com/citusdata/packaging/pull/473

PR #351, opened 2019-10-10
Update docker images

hanefi:
## Changes
- Add PG 12 support
- Introduce debian buster in build matrix

---------

## Deviating distro packaging steps
If we run `update_dockerfiles` we will start to see some extra steps in two distros as they deviated from others:
- Oracle Linux 7: added some extra steps because `From January 2019 the previous method of refreshing the Yum repositories has altered` (See https://oracle-base.com/articles/linux/download-the-latest-oracle-linux-repo-file#oracle-linux-7-updated)
- Debian Buster: Needed one extra `zlib1g-dev` package when releasing the latest `pg_auto_failover` version

-----------

## Distros without support (yet)
- el-ol/8
	- No `ol8_software_collections1` `ol8_developer` `ol8_developer_EPEL` repositories available yet. Needs a lot more changes to dockerfiles to install the packaging dependencies
- Ubuntu Trusty EOLd. No PGDG repositories for this distro

## Other problems:
- We used to install latest Citus packages to get header files. I do not understand the logic behind this, and I did not want to install PG11 in an image that will create PG12 packages.
- `make_pg_buildext_parallel.patch` no longer can be applied to the latest `pg_buildext` scripts. We need to update the patch to continue using it.
- We may need to check if the gpg key for the debian builds have expired. A comment suggests that it did : `pub   4096R/ACCC4CF8 2011-10-13 [expires: 2019-07-02]` in https://github.com/citusdata/packaging/pull/351/files/c3bae51258c5d89df5b4c6aaa15b8ae3ad9a5490#diff-0199464a84de2fb83ab569432ffa676fR25

PR #465, opened 2020-05-04
Fix the templates and introduce new distros

hanefi:
This PR aims to (i) make the templates more generic, and hence get them to create all the images once again (ii) introduce 2 new distros
- Oracle Linux 8
- Ubuntu Focal (20.04 LTS)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,10 @@ env:
     - TARGET_PLATFORM=debian,buster
     - TARGET_PLATFORM=debian,jessie
     - TARGET_PLATFORM=debian,stretch
+    - TARGET_PLATFORM=oraclelinux,8
     - TARGET_PLATFORM=oraclelinux,7
     - TARGET_PLATFORM=oraclelinux,6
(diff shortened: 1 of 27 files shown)


JelteF (approved):
Looks great. Can you comment out line 24 in `.travis.yml` so we actually check that `update_dockerfiles` continues working?

hanefi:
Currently OL8 tests fail because we can not install PackageCloud repositories in the packager. PackageCloud does not yet have any packages for OL8, and does not allow such a repo installation as the repo is not there yet.

In short we have a circular dependency.
- To create a package for a distro, we need a packaging image.
- To have a packaging image, we need to install packagecloud repo
- To have a PackageCloud repo for a new distro, you need to have packages.

~I will break this circular dependency by removing the packagecloud installation steps as advised in here: https://github.com/citusdata/packaging/pull/351#discussion_r339422037~

Edit: `pg_aut_failover_enterprise` might depend on the package cloud installation steps. I do not want to remove those lines. Instead I contacted PackageCloud support

PR #466, opened 2020-05-06
Fix issues on travis with docker images

JelteF:
1. Don't install citus packages anymore. They were only needed for the
   shard_rebalancer, which is now part of citus.
2. Put the ipv6 fix for gpg in the temporary gpg home directory too.


hanefi (approved):
LGTM

PR #468, opened 2020-05-06
Update PG10 images

JelteF:
These are needed for pg_auto_failover

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,5 +21,5 @@ before_install:
   - echo $TARGET_PLATFORM > os-list.csv
 install: true
 script:
-  - ./update_dockerfiles && git diff --exit-code dockerfiles
+  - ./update_dockerfiles && git add --intent-to-add dockerfiles && git diff --exit-code dockerfiles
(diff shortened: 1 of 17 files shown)


hanefi:
I guess ol/8 is missing. Did you intentionally left it out?

Also when I run `./update_dockerfiles` a new `dockerfiles/oraclelinux-8-pg10/` directory dirties my git working directory. This should break our tests.

hanefi:
> I guess ol/8 is missing. Did you intentionally left it out?
> 
> Also when I run `./update_dockerfiles` a new `dockerfiles/oraclelinux-8-pg10/` directory dirties my git working directory. This should break our tests.

Apparently `git diff --exit-code dockerfiles` returns 0 when there are untracked files. What do you think of using the following instead:

```bash
./update_dockerfiles && [[ -n $(git status -s dockerfiles) ]]
```

JelteF:
No I did not do this intentionally. They should be added now and I updated the command to fail our tests. It still uses git diff, since it's quite nice to see what changed too and not just which files.

PR #469, opened 2020-05-06
Bump pgautofailover all version to 1.3.1

JelteF:


debian/changelog
```diff
@@ -1,3 +1,16 @@
+pg-auto-failover (1.3.citus-1) stable; urgency=low
+
+  * Fix systemd integration [#212]
```
hanefi: We have empty lines between changelog items.

pg-auto-failover.spec
```diff
@@ -9,11 +9,11 @@ Summary:	Postgres extension for automated failover and high-availability
 Name:		%{sname}%{?pkginfix}_%{pgmajorversion}
 Provides:	%{sname}_%{pgmajorversion}
 Conflicts:	%{sname}_%{pgmajorversion}
-Version:	1.2
+Version:	1.3.citus
```
hanefi: Is this actually intended? Do we want this for consistency between packages for different extensions?

hanefi (changes requested):

pg-auto-failover.spec
JelteF: Yes I guess the reason is that it's clear where the package comes from, i.e. built by us not pgdg.

debian/changelog
JelteF: This is automatically filled by `update_os_package.pl`. This is simply the way the changelog is formatted on the pg_auto_failover repo.

DimCitus: The debian changeling is supposed to contain entries relevant to the debian packaging, not to the software being packaged. So you could have entries about the improvements you made in the debian packaging, or just one line mentioning it's an official release of the software...

pg-auto-failover.spec
```diff
@@ -9,11 +9,11 @@ Summary:	Postgres extension for automated failover and high-availability
 Name:		%{sname}%{?pkginfix}_%{pgmajorversion}
 Provides:	%{sname}_%{pgmajorversion}
 Conflicts:	%{sname}_%{pgmajorversion}
-Version:	1.2
+Version:	1.3.1.citus
```
DimCitus: Is there a reason to suffix with Citus here?

debian/changelog
DimCitus: See https://www.debian.org/doc/debian-policy/ch-source.html#debian-changelog-debian-changelog for more information and details.

pg-auto-failover.spec
DimCitus: Yes but it's confusing, is it the package for pg_auto_failover with Citus support?

pg-auto-failover.spec
JelteF: Yes, the reason is to easily differentiate our packages from ones installed through other means, i.e. pgdg or OS packages.

pg-auto-failover.spec
JelteF: That's true, I removed it now

debian/changelog
hanefi: I suggest we have the usual single entry here:
```
  * Official release of 1.3.1
```

We will be missing an entry for 1.3 but I guess this is the way it is here, and we are missing several other changelog sections as well.

pg-auto-failover.spec
```diff
@@ -82,6 +82,9 @@ PATH=%{pginstdir}/bin:$PATH
 
 
 %changelog
+* Thu May 7 2020 - Jelte Fennema <Jelte.Fennema@microsoft.com> 1.3.1-1
+- Official 1.3.1 release of pg_auto_failover
```
hanefi, line 86: I would have wanted to see a consistent item in here such as:
```
  - Official release for 1.3.1
```

debian/changelog
```diff
@@ -1,3 +1,9 @@
+pg-auto-failover (1.3.1-1) stable; urgency=low
+
+  * Official 1.3.1 release of pg_auto_failover
```
hanefi, line 3: I would have wanted to see a consistent item in here such as:
```
  - Official release for 1.3.1
```

hanefi (approved):
After 2 simple changes in changelog items for consistency between all sections, we can merge this

pg-auto-failover.spec
JelteF: I think it's better with the name of the tool in there too. 
I've modified `update_os_package.pl` to do this automatically for new versions:
https://github.com/citusdata/tools/pull/69/files#diff-99ba4ed771cad68c7bae27b53c0c4aa3R91

PR #475, opened 2020-05-06
Add ccache to el/8 and ol/8

JelteF:
The postgres make files reference ccache directly


hanefi (approved):
LGTM

hanefi:
I wonder why postgresql12-devel package does not depend on ccache. ~That does not really make sense.~

I guess ccache is neither in official centos/redhat repositories nor pgdg. As it is on a separate repository (EPEL) it starts to make some sense. Still I do not get RedHat packaging repositories

PR #476, opened 2020-05-06
Support different pkgname for rpm and deb

JelteF:


onurctirtir:
We can merge the pr as long as you tested the packaging for any of the distros with those changes. Let's also wait that travis/push job to be complete.

PR #499, opened 2020-07-22
Bump Citus PGXN to 9.3.4

onurctirtir:


META.json
```diff
@@ -2,15 +2,15 @@
...
    "abstract": "Scalable PostgreSQL for real-time workloads",
    "description": "Citus horizontally scales PostgreSQL across commodity servers using sharding and replication. Its query engine parallelizes incoming SQL queries across these servers to enable real-time responses on large datasets.",
-   "version": "9.3.0",
+   "version": "9.3.4",
    "maintainer": "\"Citus Data\" <packaging@citusdata.com>",
    "license": "agpl_3",
    "provides": {
       "citus": {
          "abstract": "Citus Distributed Database",
          "file": "README.md",
```
onurctirtir:
```suggestion
         "file": "src/backend/distributed/sql/citus--9.3-2--9.4-1.sql"
```

META.json
```diff
@@ -2,15 +2,15 @@
...
    "description": "Citus horizontally scales PostgreSQL across commodity servers using sharding and replication. Its query engine parallelizes incoming SQL queries across these servers to enable real-time responses on large datasets.",
-   "version": "9.3.0",
+   "version": "9.3.4",
    "maintainer": "\"Citus Data\" <packaging@citusdata.com>",
    "license": "agpl_3",
    "provides": {
       "citus": {
          "abstract": "Citus Distributed Database",
-         "file": "README.md",
+         "file": "src/backend/distributed/sql/citus--9.3-2--9.4-1.sql"
```
onurctirtir:
```suggestion
         "file": "src/backend/distributed/sql/citus--9.3-2--9.4-1.sql",
```

META.json
onurctirtir: lets use migration script, or should I put citus.so ? @hanefi

META.json
onurctirtir: missing comma

META.json
hanefi: `src/backend/distributed/sql/citus--9.3-2--9.4-1.sql` should not be present in this package.

I think it's better to keep maintanence costs low. Let's have a dummy value here that does not need to be updated across versions. 

I wonder why we have `RREADME.md` here. How does `citus.so` sound?

onurctirtir: sounds good, the only thing I'm not sure is where the `citus.so` is placed in directory hierarchy, would directly writing ``citus.so` work ?

hanefi: In the past, we made a typo and the package still got created. So I don't think there is any checks for validity of file paths here

onurctirtir: done

hanefi (approved):
:shipit:

PR #509, opened 2020-08-07
Use microsoft email for releases

JelteF:


scripts/determine_email
```diff
@@ -12,7 +12,7 @@ failure=1
 email=$(curl -sf https://api.github.com/user | jq -r '.email // empty')
 
 # find the first verified Citus Data email
```
hanefi: Can we update the comment as well?
```suggestion
# find the first verified Microsoft email
```

PR #510, opened 2020-08-07
Update docker files to use microsoft email for releases

onurctirtir:
Run update_docker for the changes done in #509, so update the docker images


hanefi:
Oracle Linux 7 repositories do not have a required package `libedit-devel`. We are aware of the issue, and it will not be a problem

PR #511, opened 2020-08-17
Fix determine_email so it recognizes bots@citusdata.com for nightly package builds

onurctirtir:
Due to the last changes (#509 & #510 ) that we made in `determine_email`, that script wasn't able to fetch the email address of the bot account as its email address was ending with @citusdata.com.

Try to do an exact match for bots@citusdata.com email if can't find Microsoft email.

This was blocking cron job to create nightly packages because determine_email couldn't determine email of our bot account.

### DIFF
--- a/dockerfiles/centos-6-pg10/scripts/determine_email
+++ b/dockerfiles/centos-6-pg10/scripts/determine_email
@@ -11,8 +11,9 @@ failure=1
 # fallback to public email
 email=$(curl -sf https://api.github.com/user | jq -r '.email // empty')
 
-# find the first verified Microsoft email
-jqfilter='map(select(.verified and (.email | test("@microsoft.com$")))) | first | .email // empty'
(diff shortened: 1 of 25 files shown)


hanefi:
You came up with a loop to allow multiple regex patterns. Arbitrary regex patterns are allowed on `jq` `test()` function, and you can make use of that.

Instead of these changes, I think this one liner makes more sense:

```diff
-jqfilter='map(select(.verified and (.email | test("@microsoft.com$")))) | first | .email // empty'
+jqfilter='map(select(.verified and (.email | test("@microsoft.com$|^bots@citusdata.com$")))) | first | .email // empty'
```

Does this make sense to you as well?

hanefi (approved):
All looks good, except for broken ol7 builds.

Let's create an issue to track the missing packages on OL7 that is a separate issue, and fix it later

PR #514, opened 2020-08-18
Enable ol7_optional_archive repo in OL/7

### LINKED ISSUE #512: Docker image that we use for OL7 packaging cannot be built
We get below error when we try to update&build the docker image that we use for OL7 packaging:

```
Error: Package: llvm5.0-devel-5.0.1-7.el7.x86_64 (ol7_developer_EPEL)
           Requires: libedit-devel
```

Currently we can't build the docker image and we cannot merge such pr's properly, but for now, we are still able to build OS packages using the existing old OL7 docker image from dockerhub.
But we should fix this as soon as possible.

onurctirtir:
Fixes https://github.com/citusdata/packaging/issues/512


hanefi (approved):
Well done :shipit:

PR #515, opened 2020-08-18
Use tools 0.7.19

onurctirtir:


hanefi (approved):
This is ok to do, however provides no real benefit.

The only difference between 0.7.19 and 0.7.8 is a hint in `citus_package`, that does not affect behavior of our CI scripts.

You do not need to burden yourself with keeping tools versions updated on all branches of this repo

PR #525, opened 2020-10-20
Pass security flags and check gcc version

SaitTalhaNisanci:
We were not passing security flags for citus community packages, which
we are for enterprise.

Also this adds the check for gcc version to make sure we are compliant
with security.

### DIFF
--- a/citus.spec
+++ b/citus.spec
@@ -34,8 +34,19 @@ commands.
 %prep
 %setup -q -n %{sname}-%{version}
 
+# Flags taken from: https://liquid.microsoft.com/Web/Object/Read/ms.security/Requirements/Microsoft.Security.SystemsADM.10203#guide
+SECURITY_CFLAGS="-fstack-protector-strong -D_FORTIFY_SOURCE=2 -O2 -z noexecstack -fpic -Wl,-z,relro -Wl,-z,now -Wformat -Wformat-security -Werror=format-security"
+
+currentgccver="$(gcc -dumpversion)"
+requiredgccver="4.8.2"
+if [ "$(printf '%s\n' "$requiredgccver" "$currentgccver" | sort -V | tail -n1)" = "$requiredgccver" ]; then
+      echo WARNING: Using slower security flags because of outdated compiler
+      SECURITY_CFLAGS="-fstack-protector-all -D_FORTIFY_SOURCE=2 -O2 -z noexecstack -fpic -Wl,-z,relro -Wl,-z,now -Wformat -Wformat-security -Werror=format-security"
+    fi
+fi
+
 %build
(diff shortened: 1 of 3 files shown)


JelteF (dismissed):

citus.spec
```diff
@@ -45,6 +45,11 @@ if [ "$(printf '%s\n' "$requiredgccver" "$currentgccver" | sort -V | tail -n1)"
     fi
 fi
 
+gccgte8=$(expr `gcc -dumpversion | cut -f1 -d.` \>= 8)
+ifeq "$(gccgte8)" "1"
```
JelteF, line 49: isn't `ifeq` a `Makefile` thing? I believe this spec file is (ba)sh.

citus.spec
```diff
@@ -34,8 +34,24 @@ commands.
...
+
+currentgccver="$(gcc -dumpversion)"
+requiredgccver="4.8.2"
+if [ "$(printf '%s\n' "$requiredgccver" "$currentgccver" | sort -V | tail -n1)" = "$requiredgccver" ]; then
+      echo WARNING: Using slower security flags because of outdated compiler
+      SECURITY_CFLAGS="-fstack-protector-all -D_FORTIFY_SOURCE=2 -O2 -z noexecstack -fpic -Wl,-z,relro -Wl,-z,now -Wformat -Wformat-security -Werror=format-security"
+    fi
+fi
+
+gccgte8=$(expr `gcc -dumpversion | cut -f1 -d.` \>= 8)
```
JelteF, line 48: I'd prefer if this used the same type of check as the one above for consistency, i.e. using `sort -V` (version sort).

SaitTalhaNisanci: Actually thinking about that a bit, we would enter that if it the version is equal to "4.8.2" but we don't want to, isn't that if check wrong in this case?

JelteF: I guess you are right, maybe a better way would be to use `head -n1` and check against currentgccver

SaitTalhaNisanci: if currentgccver and requiredgccver are the same, then what is the difference between head -n1 or tail -n1, or checking against `currentgccver` or `requiredgccver`. I would say that we can switch to `gte` structure here

SaitTalhaNisanci:
```bash
if ! [ "$(printf '%s\n' "$requiredgccver" "$currentgccver" | sort -V | head -n1)" = "$requiredgccver" ]; then
      echo WARNING: Using slower security flags because of outdated compiler
fi
```

or something like the above maybe

JelteF: yes that seems good, but maybe use `!=`  instead of `! [ ... = ... ]`

And then use 
```bash
if [ "$(printf '%s\n' "8.0.0" "$currentgccver" | sort -V | head -n1)" = "8.0.0" ]; then
      # add extra security flag
fi
```

PR #528, opened 2020-10-21
Fail in determine_name if any commands fails

SaitTalhaNisanci:
I didn't have `jq` installed and it would not fail because `-e` option wasn't set.


JelteF (approved):

hanefi:
That script is not expected to be run on your machine. They are intended for use only when building docker images.

However, if for some reason we break `jq` installation in a docker image, `-e` option can successfully break our builds and provide value.

hanefi:
You also need to update the scripts in many places by running `./update_dockerfiles`.

See https://github.com/citusdata/packaging#updating-dockerfiles

PR #532, opened 2020-11-02
Build package builder docker images with pg13

onurctirtir:
* e67b522f2261962a8e39a2aec20c7ea0612d7fd6
Applies base changes for pg13 support.

* 4b2929fbeda21e7bdf14a1dbd81e6e1cf52bfd7f
As CentOS/OracleLinux 6 doesn't have essential pg13 packages, disable updating "package builder docker image" for those distros, and disable citus package building as well.
Even if job for debian/jessie didn't fail, it doesn't have essential pg13 packages yet like CentOS/OracleLinux 6, so we skip it as well. (see http://apt.postgresql.org/pub/repos/apt/dists/jessie-pgdg/)

* 622250f2d9b3cb065b230160a7c54ef10a162861
Run update_dockerfiles to apply base changes.

----

I will build some of the docker images locally as well, and will test if packaging works fine with pg13 for some random distros.


hanefi:
I suggest you remove the new folders for el6 & PG13 that do not build because of missing packages in PGDG repos.

Also I suggest you keep the changes for making exceptions for EL6 in a separate commit, so that we can revert it once the PGDG packages are there.

onurctirtir:
> I suggest you remove the new folders for el6 & PG13 that do not build because of missing packages in PGDG repos.

But then, ci wouldn't pass ?


> Also I suggest you keep the changes for making exceptions for EL6 in a separate commit, so that we can revert it once the PGDG packages are there.

Agree 6e4b608

hanefi:
> > I suggest you remove the new folders for el6 & PG13 that do not build because of missing packages in PGDG repos.
> 
> But then, ci wouldn't pass ?

If you add some logic for excluding EL6&PG13 combination in the lines below, then CI would pass I think.
https://github.com/citusdata/packaging/blob/6e4b608a9d41d2efdce1b70f386ee85a89965fb4/update_dockerfiles#L92-L94

.travis.yml
```diff
@@ -4,15 +4,19 @@ language: perl
...
+    # So skip building docker images for them.
+    # If you build docker images locally you need to edit os-list.csv manually !
+
     - TARGET_PLATFORM=centos,8
     - TARGET_PLATFORM=centos,7
-    - TARGET_PLATFORM=centos,6
+    #- TARGET_PLATFORM=centos,6
     - TARGET_PLATFORM=debian,buster
-    - TARGET_PLATFORM=debian,jessie
+    #- TARGET_PLATFORM=debian,jessie
```
hanefi: Debian Jessie EOL'd on June 30, 2020. We can safely remove this line.

build_packages
```diff
@@ -45,13 +45,19 @@ while read -r line; do
     mkdir -p "${outputdir}"
 
     if [[ "${os}" = 'debian' ]] || [[ "${os}" = 'ubuntu' ]]; then
+        if [[ "${os}" = 'debian' ]] && [[ "${release}" = 'jessie' ]]; then
```
hanefi, line 48: Debian Jessie EOL'd. We can remove these changes

hanefi (changes requested):

.travis.yml
hanefi: On second thought, we are able to create PG11 and PG12 packages for Debian Jessie.

Should we keep creating packages for Jessie only for a limited set of PG versions?

onurctirtir: >Should we keep creating packages for Jessie only for a limited set of PG versions?

I think yes

hanefi (changes requested):
I checked all CI builds, and they are ok except Debian Jessie is missing.

If we do not drop support for Debian Jessie, we should still keep it in CI. You can modify scripts used in CI to exclude PG13 builds for Debian Jessie, I guess.

update_dockerfiles
```diff
@@ -90,6 +90,12 @@ while read -r line; do
         # redhat variants need a Dockerfile for each PostgreSQL version
         IFS=' '
         for pgversion in ${pgversions}; do
+            if { [[ "${os}" = 'centos' ]] || [[ "${os}" = 'oraclelinux' ]]; } && \
+                 [[ "${release}" = '6' ]] && [[ "${pgversion}" = '13' ]]; then
+                # CentOS and OracleLinux 6 doesn't have pg13 packages yet.
+                # So skip building docker images for them.
+                continue
+            fi
```
hanefi: nit, trailing whitespace

```suggestion
            fi
```

onurctirtir: addressed

hanefi (approved):
:shipit:

Issue #536, opened 2020-11-06
After updating docker images, debian package builds would fail with configure error


onurctirtir:
I locally built new package builder docker images and I was testing to produce packages via those new docker images. 
However, in debian, we have the following error:

>citus_package -p=debian/stretch local release-9.5 2>&1 | tee -a citus_package.log

```bash
dh clean
   dh_testdir
   debian/rules override_dh_auto_clean
make[1]: Entering directory '/9.5.0.citus~pre.20201106.fee46d6'
pg_buildext clean build-%v
rm -rf build-*/
make[2]: Entering directory '/9.5.0.citus~pre.20201106.fee46d6'
Makefile:9: *** ./configure needs to be run before compiling Citus.  Stop
make[2]: Leaving directory '/9.5.0.citus~pre.20201106.fee46d6'
debian/rules:9: recipe for target 'override_dh_auto_clean' failed
make[1]: *** [override_dh_auto_clean] Error 2
make[1]: Leaving directory '/9.5.0.citus~pre.20201106.fee46d6'
debian/rules:21: recipe for target 'clean' failed
make: *** [clean] Error 2
dpkg-buildpackage: error: fakeroot debian/rules clean gave error exit status 2
debuild: fatal error at line 1116:
dpkg-buildpackage -rfakeroot -us -uc -B failed
docker run failed. see output for details.
```

Looking into `Makefile:9: *** ./configure needs to be run before compiling Citus.  Stop` part,
it seems that `clean` step [here](https://github.com/citusdata/packaging/blob/all-enterprise/debian/rules#L10) somehow breaks execution of `./configure`

hanefi:
I think we solved this issue for all projects.

PR #537, opened 2020-11-06
Remove clean step from debian rules

onurctirtir:
Fix for #536 for citus/community


hanefi:
Do we need to do this for other extensions as well?

hanefi (approved):
:shipit:

onurctirtir:
> Do we need to do this for other extensions as well?

Probably yes, so I won't close  the issue  #536

PR #542, opened 2020-11-11
Bump Citus version to 9.5.0 & Enable PG13

onurctirtir:


hanefi (approved):
All the expected packages seem to be fine.

We do not create PG13 packages for EL6, OL6 and Debian Jessie, as these distros have missing dependencies from PGDG.
:shipit:

PR #545, opened 2020-11-12
Update deb and rpm scripts to use PG13

onurctirtir:
Tested 9bfde1b5f5b667d74e26435ff3697586d25b78b1 on several distros for pg13, works fine.
For the distros that are not having pg13 packages, had some issues and fixed them in other commits

### DIFF
--- a/community-nightlies/deb.sh
+++ b/community-nightlies/deb.sh
@@ -56,9 +56,9 @@ curl_check ()
 
 pgdg_check ()
 {
-  echo "Checking for postgresql-12..."
-  if apt-cache show postgresql-12 &> /dev/null; then
-    echo "Detected postgresql-12..."
+  echo "Checking for postgresql-13..."
+  if apt-cache show postgresql-13 &> /dev/null; then
+    echo "Detected postgresql-13..."
   else
     pgdg_list='/etc/apt/sources.list.d/pgdg.list'
     pgdg_source_path="deb http://apt.postgresql.org/pub/repos/apt/ ${codename}-pgdg main"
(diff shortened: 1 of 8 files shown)


onurctirtir:
Tested on Ubuntu 18.04, works fine

community-nightlies/deb.sh
```diff
@@ -87,6 +82,16 @@ pgdg_check ()
     # import the gpg key
     curl -L "${pgdg_key_url}" 2> /dev/null | apt-key add - &>/dev/null
     echo "done."
+
+    echo -n "Running apt-get update... "
+    apt-get update &> /dev/null
+    echo "done."
+
+    if ! apt-cache show postgresql-13 &> /dev/null; then
+      echo "PGDG doesn't postgresql-13 package for operating system"
```
hanefi:
```suggestion
      echo "PGDG repositories don't have postgresql-13 package for your operating system"
      echo
```

onurctirtir: addressed

community-nightlies/rpm96.sh
```diff
@@ -39,6 +39,13 @@ pgdg_check ()
     echo -n "Installing pgdg96 repo... "
 
     yum install -d0 -e0 -y "${repo_url}"
+
+    if ! yum info -q postgresql13-server &> /dev/null; then
```
hanefi: `(deb|rpm)96.sh` scripts are for installing PG 9.6 and supported extensions.

AFAIK, they are outdated and can be removed. It does not makes sense to me that we are checking for PG13 packages for a script that aims to install PG9.6 repos.

Let's keep these files as is, and maybe remove them in a seperate PR.

hanefi (changes requested):

hanefi (approved):

PR #549, opened 2020-11-23
Bump pg_auto_failover enterprise version to 1.3.2

onurctirtir:


DimCitus (approved):

hanefi:
Interesting error on ubuntu Xenial build https://travis-ci.org/github/citusdata/packaging/jobs/745380240
```
E: pg-auto-failover-cli-enterprise-1.3: versioned-provides pg-auto-failover-cli-enterprise (= 1.3.2-1)
```

DimCitus:
> Interesting error on ubuntu Xenial build https://travis-ci.org/github/citusdata/packaging/jobs/745380240
> 
> ```
> E: pg-auto-failover-cli-enterprise-1.3: versioned-provides pg-auto-failover-cli-enterprise (= 1.3.2-1)
> ```

Given https://lintian.debian.org/tags/invalid-versioned-provides.html ; yeah, “interesting”.

onurctirtir:
I see that we use the proper operator, so couldn't understand why we have this error message
https://github.com/citusdata/packaging/blob/all-pgautofailover-enterprise/debian/control.in#L13

DimCitus:
Could it be because we include the debian version number in our version number? Is it possible to try with `(= 1.3.2)`, that is only with the package version number, without the debian package revision?

onurctirtir:
> Could it be because we include the debian version number in our version number? Is it possible to try with `(= 1.3.2)`, that is only with the package version number, without the debian package revision?

Yeah, it could be related with `-` version, but as I talked with @JelteF, it doesn't seem to be a problem, it's there for a while for some old debian based distros.

So, now are we good to merge @DimCitus , @hanefi  ?

hanefi (approved):
LGTM
:shipit:

DimCitus:
If you could install the resulting package without trouble, then I'm good. We need to make sure the Lintian check won't activate and prevent installation of the package on the users/customer systems.

JelteF:
For context it's not `invalid-versioned-provides`, but the now removed `versioned-provides`.
Error description: http://people.ubuntuwire.org/~broder/lintian/quantal/tags/versioned-provides.html
Removal of the lint: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=800438

PR #552, opened 2020-11-24
Bump Citus Enterprise version to 9.4.3

onurctirtir:


hanefi (approved):
The columnar store `.bc` files on master branch require different spec `%files` entries for nightly packages to work.

I'm ok with stopping nightly image creation to unblock 9.4.3 patch release

:shipit:

debian/changelog
```diff
@@ -1,3 +1,11 @@
+citus-enterprise (9.4.3.citus-1) stable; urgency=low
+
+  * Enables PostgreSQL's parallel queries on EXPLAIN ANALYZE
+
+  * Fixes a bug that triggers subplan executions unnecessarily with cursors
+
+ -- Onur Tirtir <Onur.Tirtir@microsoft.com>  Tue, 24 Nov 2020 11:22:0 +0000
```
hanefi:
```suggestion
 -- Onur Tirtir <Onur.Tirtir@microsoft.com>  Tue, 24 Nov 2020 11:22:00 +0000
```

This will break builds on some distros

hanefi (changes requested):
One minor timestamp formatting issue. Looks good otherwise

debian/changelog
onurctirtir: yes, was just about to do that :)

PR #554, opened 2020-11-24
citus-community/Include .bc files for columnar as well

### LINKED ISSUE #547: Nigthly builds failed for rpm>6 distros after merging cstore to citus
I will look into that soon
https://travis-ci.org/github/citusdata/packaging/builds/744401514

edit:
Probably will also affect regular releases

onurctirtir:
At d2e7dfd: Fix #547 for citus/community
At a43803c: Revert 86b872b to enable nightly builds again

To explain what d2e7dfd does:
* We need to add .bc files for columnar for Citus versions >= 10.0
* We need to do that conditionally, depending on if columnar directory exists
* Directly including `%{pginstdir}/lib/bitcode/columnar/*.bc` breaks builds for Citus versions < 10.0
* So we need to write that columnar path conditionally to a file list
* But this file can't be empty, otherwise build is again broken
* So include other installation files in that files list, and optionally include columnar dir

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_install:
   - sudo make -C tools install
   - export PACKAGING_SECRET_KEY="$(openssl aes-256-cbc -K $encrypted_a39e3f1ee8ba_key -iv $encrypted_a39e3f1ee8ba_iv -in signing_key.asc.enc -d | base64)"
 install: true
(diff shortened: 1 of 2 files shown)


onurctirtir:
Patch releases don't have columnar, so this solution does not solve the issue for all releases

hanefi (approved):
:shipit:

PR #557, opened 2020-11-24
Revert pgxn 9.4.3

onurctirtir:
https://github.com/citusdata/packaging/pull/556 bumps pgxn to 9.4.3.
But in https://travis-ci.com/github/citusdata/packaging/builds/203814764, we see following error:
>PGXN already contains citus-9.4.3.zip!
The command "release_pgxn" exited with 0.

It's probably because pgxn doesn't accept artifacts for older versions and trying to make it work wouldn't add any value, so revert that pr


hanefi:
>It's probably because pgxn doesn't accept artifacts for older versions 

This is indeed the case. Backport releases are rejected by pgxn.

hanefi (approved):

PR #559, opened 2020-11-26
Bump redhat hll to 2.15

hanefi:


onurctirtir (approved):
LGTM

Could have seperate commits like:
* Extend supported distros with el/6, ol/6 & el/8
* Use tools v0.7.21
* Remove old pg versions
* Add support for pg13
* Bump redhat hll to 2.15

In general, doing things in seperate pr's, -or at least in seperate commits- would make the things easier for the next person that would do the same thing in future.

PR #560, opened 2020-11-26
Bump debian hll to 2.15

hanefi:


onurctirtir (commented):
1- Compared the the changes that we made just to bump hll with https://github.com/citusdata/packaging/pull/368/, looks good

2- Tests also look good

------------

3- I see that we had no such issues as mentioned in https://github.com/citusdata/packaging/issues/536, but should we do something similar to https://github.com/citusdata/packaging/pull/538 to be on the safe side ?

4- Could be better if we can have commits split into different pieces, that way we can track what we did in this pr in future.
For https://github.com/citusdata/packaging/pull/560/commits/d7b495a1bb21770b219b74a23eac207f1eb7f224, we can have three commits:
* Bump hll to 2.15
* Add ubuntu/focal
* Add pg13

For https://github.com/citusdata/packaging/pull/560/commits/dc61fd66997c5c71a52e32a6efde5607726eca89:
* Use tools x.y
* Remove older pg ..

hanefi:
> 3- I see that we had no such issues as mentioned in #536, but should we do something similar to #538 to be on the safe side ?

We do not have any autoconf scripts in this project, so I don't think this issue applies here.

> 4- Could be better if we can have commits split into different pieces, that way we can track what we did in this pr in future.


Good advice 👍  I will split my changes into smaller commits

onurctirtir (approved):
Thanks for splitting into logical pieces, please use "merge" when merging the pr

onurctirtir:
We forgot to remove debian/jessie from supported OS'es, but it's not much important :)

PR #561, opened 2020-11-27
Install libstdc++ for hll packaging

hanefi:
I updated images for only rpm based distros that fail to build hll extension due to missing libstdc++ dependencies.


hanefi:
~Currently~ Was blocked on #564 that fixes debian images

onurctirtir:
> Currently blocked on #561 that fixes debian images

isn't it the same pr :)

onurctirtir (approved):
A side question, why don't we need to install that library in debian distros ?

hanefi:
> A side question, why don't we need to install that library in debian distros ?

Tough question. I do not know the answer.

If I were to guess, I would say that dependencies on rpm packages sometimes miss out some crucial ones on purpose. For example if a development package has some dependencies from EPEL repos, they may leave them out to not force users to install EPEL repos.

PR #567, opened 2020-11-30
Bump debian TopN to 2.3.1

hanefi:
We need to fix several issues for a clean release. See #560 for a similar PR


onurctirtir (approved):
For readers of this pr (https://github.com/citusdata/packaging/pull/567/commits/392e6084ee318da86b3d00e864d7742d24df9634):

from https://manpages.debian.org/experimental/postgresql-server-dev-all/pg_buildext.1.en.html:
>loop [src-dir] package-pattern
As a variant to calling build and install separately for VPATH builds, loop over the supported PostgreSQL versions in the top source directory. This should be used if the package does not support VPATH builds. As it also invokes make install, it should be placed were installation happens in debian/rules, rather than where build would normally be called.

onurctirtir:
We forgot to remove debian/jessie from supported OS'es, but it's not much important :)

PR #578, opened 2020-12-22
updates package cloud repo secure key

gurkanindibay:

PR #579, opened 2020-12-23
Fix pg-cron builds and add ubuntu/focal support

gurkanindibay:
* Updates packagecloud & github secure keys
* Fixes a build issue due to invalid date in changelog
* Removes debian/wheezy support
* Removes build dependencies that don't exist anymore

Finally,
* Adds ubuntu/focal support
* Also, to support ubuntu/focal, updates tools version to 0.7.20

Closes #578


onurctirtir (approved):
As we don't have changes in individual commits, please follow squash strategy for when merging pr. Thanks

PR #582, opened 2021-01-05
Bump Citus version to 9.4.4

gurkanindibay:


onurctirtir (approved):
Thanks for making changes in seperate commits.

A small improvement note for next time:
It could be slightly better if we separated the changes removing pg13 from supported versions to see why/when we did that clearly.

https://github.com/citusdata/packaging/pull/582/commits/bfbee3340e917ff014a4dbfffecb10cfb9b6785b

PR #585, opened 2021-01-07
Add lz4 and zstd support for all linux distros on docker to fix nightly build problem

### LINKED ISSUE #581: Nightly citus builds are broken due to missing dependency to lz4
It's because we introduced dependency to `lz4` in Citus repo.
Kind of an expected issue: https://github.com/citusdata/packaging/runs/1608461061
We should fix this asap. This will also break Citus 10.0 release.

gurkanindibay:


dockerfiles/centos-7-pg10/Dockerfile
```diff
@@ -39,6 +39,10 @@ RUN ( [[ -z "epel-release centos-release-scl-rh" ]] || yum install -y epel-relea
         rpmlint \
         spectool \
         tar \
+        lz4 \
+        lz4-devel \
+        libzstd \
+        libzstd-devel \
```
onurctirtir: Why are we installing both x & x-devel packages ? Also, ol7 seems to be broken, can you check why ?

gurkanindibay: Both files required. I have Niles checked.

onurctirtir (approved):

dockerfiles/centos-7-pg10/Dockerfile
thanodnl: strictly speaking the [devel packages depend](https://packages.debian.org/buster/liblz4-dev) on the non devel packages, so they are not strictly required.

However I think it will be more readable between the container and the packages to include them both.

dockerfiles/centos-7-pg10/Dockerfile
```diff
@@ -19,6 +19,18 @@ RUN [[ centos != oraclelinux ]] || [[ 7 != 7 ]] || ( \
...
+
+RUN if [[ centos   == oraclelinux ]] && [[ 7 == 7 ]]; then yum install -y wget \
+        && wget http://mirror.centos.org/centos/7/os/x86_64/Packages/lz4-devel-1.8.3-1.el7.x86_64.rpm \
+        && rpm -ivh lz4-devel-1.8.3-1.el7.x86_64.rpm ; \
+        elif [[ centos   == oraclelinux ]] && [[ 7 == 6 ]]; then yum install -y wget \
+        && wget https://cbs.centos.org/kojifiles/packages/lz4/r131/1.el6/x86_64/lz4-r131-1.el6.x86_64.rpm \
+        && rpm -ivh lz4-r131-1.el6.x86_64.rpm \
+        && wget https://cbs.centos.org/kojifiles/packages/lz4/r131/1.el6/x86_64/lz4-devel-r131-1.el6.x86_64.rpm \
+        && rpm -ivh lz4-devel-r131-1.el6.x86_64.rpm;  \
+        else yum install -y lz4 lz4-devel; fi
```
thanodnl, line 34: could use a comment why we are installing it like this instead of adding the packages to the statement below. I guess not all centos verions do support the packages? Could this become an issue for our releases?

thanodnl (approved):
Looks good.

I think it is worth adding a small comment explaining the ifstatement and the wget's for future you and others to know what and why we are doing it like that.

dockerfiles/centos-7-pg10/Dockerfile
gurkanindibay: Ok I'm adding comment

PR #594, opened 2021-02-05
Bump pg_auto_failover enterprise version to 1.4.2

gurkanindibay:


hanefi (approved):
We have a new warning that is about a new not accessible documentation page that is not linked in a table of contents. This is not a blocker for us as we ship documentation in other forms where this is not an issue for the packages.

PR #595, opened 2021-02-11
Bump pg_auto_failover version to 1.4.2

gurkanindibay:


debian/changelog
```diff
@@ -1,3 +1,9 @@
+pg-auto-failover (1.4.2-1) stable; urgency=low
```
hanefi:
```suggestion
pg-auto-failover (1.4.2-2) stable; urgency=low
```

Let's bump the fancy version. Maybe we can add a changelog item as well

debian/changelog
```diff
@@ -1,3 +1,9 @@
...
+
+  * Official 1.4.2 release of pg_auto_failover
```
hanefi:
```suggestion
  * Official 1.4.2 release of pg_auto_failover
  * Adds the missing dependency postgresqlXX-contrib
```

pg-auto-failover.spec
```diff
@@ -82,6 +82,9 @@ PATH=%{pginstdir}/bin:$PATH
 
 
 %changelog
+* Thu Feb 11 2021 - Gurkan Indibay <gindibay@microsoft.com> 1.4.2-1
```
hanefi:
```suggestion
* Thu Feb 11 2021 - Gurkan Indibay <gindibay@microsoft.com> 1.4.2-2
```

pg-auto-failover.spec
```diff
@@ -82,6 +82,9 @@ PATH=%{pginstdir}/bin:$PATH
...
+- Official 1.4.2 release of pg_auto_failover
```
hanefi, line 86: If you decide to add a changelog item, please place the same entry here as well.

hanefi (changes requested):
Please remove Debian based distros from the CI build matrix in a separate commit. On future releases we can revert that commit to get back to the original list of distros.

debian/changelog
gurkanindibay: We will not release for  debiam so I think we don't need to change the changelog

debian/changelog
gurkanindibay: I removed the unnecessary entry because there is no need to release for debian

pg-auto-failover.spec
gurkanindibay: Fixed it

hanefi (approved):

Issue #491, opened 2020-05-27
Strip citus.so .comment .note sections


hanefi:
We have some `unstripped-binary-or-object` [warnings in OL/7](https://travis-ci.org/github/citusdata/packaging/jobs/691635525#L574), and possibly for some other distros as well.

```
citus-enterprise93_11.x86_64: W: unstripped-binary-or-object /usr/pgsql-11/lib/citus.so
```

Debian-policy includes [a section](https://www.debian.org/doc/debian-policy/ch-files.html#libraries) about this:
> All installed shared libraries should be stripped with
> 
> ```
> strip --strip-unneeded --remove-section=.comment --remove-section=.note your-lib
> ```
>
> (The option --strip-unneeded makes strip remove only the symbols which aren’t needed for relocation processing.) Shared libraries can function perfectly well when stripped, since the symbols for dynamic linking are in a separate part of the ELF object file.
> 
> Note that under some circumstances it may be useful to install a shared library unstripped, for example when building a separate package to support debugging. The debhelper dh_strip` tool can create such packages automatically.
>

onurctirtir:
Possibly already addressed.

PR #598, opened 2021-02-16
Bump Citus Enterprise version to 10.0.0

gurkanindibay:


onurctirtir (approved):
I checked ci jobs and left some notes below. Please you also do check those jobs before merging.

-----

* We still see https://github.com/citusdata/citus-enterprise/issues/516 but we are okay with that for  10.0.0 release.

Other than that, not related with this pr but let me document some other things that I'm seeing over and over again in packaging jobs:
* Soon we should fix https://github.com/citusdata/packaging/issues/491 since we still see `unstripped-binary-or-object` warning
Currently we behave it as if it's a `false positive` but we wouldn't actually know if we mistakenly left some debug symbols in `citus.so` or not.
* We're getting below warning in rpm builds:
`warning: line 344: multiple %files for package 'citus-enterprise_`.
This warning is due to the way that we include `.bc` files.
However, I performed regression tests with centos/7 package and all looks fine

PR #606, opened 2021-02-22
add event type check on travis to seperate triggering cron and manually triggered events

gurkanindibay:


.travis.yml
```diff
@@ -30,7 +30,14 @@ before_install:
...
 install: true
-script: build_new_release && build_new_nightly
+script: |
+  if [ ${TRAVIS_EVENT_TYPE} == 'push' ] || [ ${TRAVIS_EVENT_TYPE} == 'pull_request' ] || [ ${TRAVIS_EVENT_TYPE} == 'api' ]; then
+  echo "Travis Event Type is ${TRAVIS_EVENT_TYPE}"
+  build_new_release
+  elif [ ${TRAVIS_EVENT_TYPE} == 'cron' ]; then
+  echo "Travis Event Type is ${TRAVIS_EVENT_TYPE}"
+  build_new_nightly
+  fi
```
onurctirtir, line 40: I think we can always run `build_new_nightly` to make sure that we won't break nightly builds when merging a pr etc., what do you think ?:

```suggestion
script: |
  if [ ${TRAVIS_EVENT_TYPE} == 'push' ] || [ ${TRAVIS_EVENT_TYPE} == 'pull_request' ] || [ ${TRAVIS_EVENT_TYPE} == 'api' ]; then
  echo "Travis Event Type is ${TRAVIS_EVENT_TYPE}"
  build_new_release
  fi 
  # run nightly job even if it's not a cron job to make sure that we won't break nightly builds
  build_new_nightly
```

gurkanindibay: The suggested solution has one side effect. If we have more than one commit in one day, we will push more than one release for one day. It'a a little bit confusing I think

onurctirtir: `build_new_nightly` would be no-op if it's run it for second time, so I think there shouldn't be such a side-effect, what do you think ?

Also, this aligns with the current approach more. That means, currently we already execute nightly job for each pr/push made to main branch (all-pgautofailoever) so we should be fine with that.

onurctirtir (approved):

.travis.yml
onurctirtir: talked personally, current approach seems better

PR #609, opened 2021-02-25
Fix broken enterprise cflags

hanefi:
It appears that pg_buildext has a bug, where a parameter that is supposed to be extra CFLAGS override current ones provided by pgxs, and we end up with missing compiler flags.

pg_buildext manpage ACTIONS section:

       build [src-dir] build-dir [extra-cflags]
           Build the extension in the build-dir directory.

There is a chain of makefiles that can potentially override CFLAGS and apparently if we want to supply extra-cflags parameter we lose original flags set by the packages.

- `include $(PGXS)` https://github.com/citusdata/citus/blob/8820541fd4e0bad7827417ee136afde57f6e834b/Makefile.global.in#L50
	- `include $(top_builddir)/src/Makefile.global` https://github.com/postgres/postgres/blob/6b40d9bdbdc9f873868b0ddecacd9a307fc8ee26/src/makefiles/pgxs.mk#L85
		- `CFLAGS = @CFLAGS@` https://github.com/postgres/postgres/blob/6b40d9bdbdc9f873868b0ddecacd9a307fc8ee26/src/Makefile.global.in#L260
	- several prepended flags. (Both := and +=) https://github.com/postgres/postgres/blob/6b40d9bdbdc9f873868b0ddecacd9a307fc8ee26/src/makefiles/pgxs.mk
- `override CFLAGS += @CFLAGS@ @CITUS_CFLAGS@` and similar changes https://github.com/citusdata/citus/blob/8820541fd4e0bad7827417ee136afde57f6e834b/Makefile.global.in#L88

The problem lies in the https://salsa.debian.org/postgresql/postgresql-common/-/blob/master/pg_buildext#L80
```
make -C $vtarget ${makefile:-} ${cflags:+CFLAGS="$cflags"} PG_CONFIG="$pgc" VPATH="$srcdir" USE_PGXS=1 $MAKEVARS || return $?
```
Where the command line argument overrides all ordinary assignments of CFLAGS variables in the makefiles. See https://www.gnu.org/software/make/manual/html_node/Overriding.html#Overriding

This PR creates a workaround until pg_buildext is patched and fixed so that we do not get blocked.

Fixes https://github.com/citusdata/citus-enterprise/issues/516
Closes https://github.com/citusdata/packaging/pull/601

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -1,10 +1,11 @@
 #!/usr/bin/make -f
 
 include /usr/share/postgresql-common/pgxs_debian_control.mk
+CFLAGS := -fstack-protector-strong -D_FORTIFY_SOURCE=2 -O2 -z noexecstack -fpic -Wl,-z,relro -Wl,-z,now -Wformat -Wformat-security -Werror=format-security
 
 override_dh_auto_build:
 	# Flags taken from: https://liquid.microsoft.com/Web/Object/Read/ms.security/Requirements/Microsoft.Security.SystemsADM.10203#guide
-	+pg_buildext build build-%v '$(CFLAGS) -fstack-protector-strong -D_FORTIFY_SOURCE=2 -O2 -z noexecstack -fpic -Wl,-z,relro -Wl,-z,now -Wformat -Wformat-security -Werror=format-security'
+	+pg_buildext build build-%v
 
 override_dh_auto_clean:
 	# This breaks override_dh_auto_configure, so disable it.
--- a/debian/rules
+++ b/debian/rules
@@ -1,7 +1,7 @@
 #!/usr/bin/make -f
 
 include /usr/share/postgresql-common/pgxs_debian_control.mk
-CFLAGS := -fstack-protector-strong -D_FORTIFY_SOURCE=2 -O2 -z noexecstack -fpic -Wl,-z,relro -Wl,-z,now -Wformat -Wformat-security -Werror=format-security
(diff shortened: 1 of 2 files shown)


hanefi:
I opened a merge request to `pg_buildext` to fix the underlying issue.

See https://salsa.debian.org/postgresql/postgresql-common/-/merge_requests/11

debian/rules
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/make -f
 
 include /usr/share/postgresql-common/pgxs_debian_control.mk
+CFLAGS := -fstack-protector-strong -D_FORTIFY_SOURCE=2 -O2 -z noexecstack -fpic -Wl,-z,relro -Wl,-z,now -Wformat -Wformat-security -Werror=format-security
```
JelteF: I think this should include any CFLAGS that are passed in by debian 
```suggestion
CFLAGS := $(CFLAGS) -fstack-protector-strong -D_FORTIFY_SOURCE=2 -O2 -z noexecstack -fpic -Wl,-z,relro -Wl,-z,now -Wformat -Wformat-security -Werror=format-security
```

JelteF (commented):
I haven't tested that this works, but it looks okay (except for my comment). It's important that someone verifies that our security flags are included in the build this way. The way I usually do it is to add a bad flag like `-doesnotexist` and verify that the compile fails.

debian/rules
hanefi: This makes sense. On my tests CFLAGS was usually empty but it does not mean that it is always be empty like that.

Thanks for the suggestion

onurctirtir (commented):
Can you also share the gcc flags that we were seeing when building (e.g) ubuntu/xenial package, before and with this pr ?

JelteF (approved):

PR #610, opened 2021-03-03
Use security flags supplied by the configure scripts

hanefi:
This PR adds support for the newly added security flags in configure scripts

### DIFF
--- a/citus.spec
+++ b/citus.spec
@@ -35,7 +35,7 @@ commands.
 %setup -q -n %{sname}-%{version}
 
 %build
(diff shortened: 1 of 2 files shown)


onurctirtir:
Talked with @metdos and merging this pr to move forward with 10.0.2 release (https://github.com/citusdata/packaging/pull/611).
Still, it would be better if @JelteF can doublecheck .

onurctirtir (approved):

hanefi:
If we can check the configure output and the actual flags during packages, it should be enough I think

JelteF (commented):
Looks good!

PR #612, opened 2021-03-04
Bump Citus Enterprise version to 10.0.2

gurkanindibay:


onurctirtir (approved):

hanefi (changes requested):
Can you add encrypted MS packages to the CI build matrix?

hanefi (approved):
>Can you add encrypted MS packages to the CI build matrix?

It appears that it is already the case. Sorry for my mistake. All looks good

PR #615, opened 2021-03-05
Add epel repos for rpm based systems

hanefi:


community-nightlies/rpm.sh
```diff
@@ -54,6 +54,37 @@ pgdg_check ()
...
+      yum install -d0 -e0 -y epel-release &> /dev/null
+    elif [ ${os} = "rhel" ] || [ ${os} = "redhatenterpriseserver" ]; then
+      if [ ${dist} = "7" ]; then
+        yum install -d0 -e0 -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-${dist}.noarch.rpm &> /dev/null
+      elif [ ${dist} = "8" ]; then
+        dnf install -d0 -e0 -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-${dist}.noarch.rpm &> /dev/null
+      fi
+    fi
+
+    epel_release_check=`rpm -qa | grep -qw epel-release`
```
gurkanindibay, line 77: we should remove -w flag (whole word search) because since we are installing from an address and a specific version, package name also includes version number as below. I got it from ubi8 by executing  rpm -qa
rpm -qa | grep  "epel-release"
epel-release-7-13.noarch

hanefi: `-w` whole world search accepts `-` characters as word boundaries. So it should be ok if the package name is `epel-release-7-13.noarch`

```
❯ echo "epel-release\nepel-release-7-13.noarch\nepel-releasewordwithoutspace" | grep -w epel-release
epel-release
epel-release-7-13.noarch
```

gurkanindibay: image

gurkanindibay: I tried it and actually can not find it

hanefi: `-q` flag means quiet and therefore you do not see any output. You can try `grep -w epel-release`.

PS: We had a chat and writing this comment for documentation purposes.

gurkanindibay: Thanks approving PR

gurkanindibay (approved):

PR #628, opened 2021-03-25
Bump pg_auto_failover version to 

gurkanindibay:


gurkanindibay:
Executed script without parameters so I closed without merge

PR #631, opened 2021-03-25
Drop postgres 11 support from packages

gurkanindibay:


pkgvars
```diff
@@ -2,6 +2,6 @@ pkgname=citus-enterprise
 pkgdesc='Citus Enterprise'
 pkglatest=10.0.3.citus-1
 releasepg=11,12,13
```
onurctirtir: let's remove 11 from here too, then I'm fine with merging
and maybe update pr title to "Drop postgres 11 support from packages"

onurctirtir (approved):

PR #632, opened 2021-03-25
Drop postgres 11 support from nightly packages

gurkanindibay:


pkgvars
```diff
@@ -2,5 +2,5 @@ pkgname=citus
 pkgdesc='Citus (Open-Source)'
 pkglatest=10.0.3.citus-1
 releasepg=11,12,13
```
onurctirtir, line 4: let's remove 11 from here too, then I'm fine with merging
and maybe update pr title to "Drop postgres 11 support from packages"

onurctirtir (approved):

pkgvars
onurctirtir: @gurkanindibay, were you planning to open another pr for that ?

PR #642, opened 2021-04-01
Add update_package_properties pipeline

gurkanindibay:


gurkanindibay:
Opened while testing update_package_properties

PR #645, opened 2021-04-02
All citus update properties pipeline

gurkanindibay:


.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,60 @@
+# This is a basic workflow to help you get started with Actions
```
hanefi: Let's remove some comments unless they provide some value. Please go over other comments as well

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,60 @@
...
+  GH_TOKEN: ${{ secrets.GH_TOKEN }}
+  PRJ_NAME: ${{ github.event.inputs.project_name }}
+  TAG_NAME: ${{ github.event.inputs.tag_name }}
+  PRJ_VER: ${{ github.event.inputs.project_version }}
+  FANCY: ${{ github.event.inputs.fancy }}
+  FANCY_VERSION_NO: ${{ github.event.inputs.fancy_version_no }}
+  MICROSOFT_EMAIL: ${{ github.event.inputs.microsoft_email }}
+  NAME: ${{ github.event.inputs.name }}
+
+# Controls when the action will run.
```
hanefi:
```suggestion
```

.github/workflows/update_package_properties.yml
hanefi:
```suggestion
```

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,60 @@
...
+on:
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+    inputs:
+      project_name:
+        description: 'The github repository name'
```
hanefi:
```suggestion
        description: 'The GitHub repository name'
```

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,60 @@
...
+      tag_name:
+        description: 'The tag name on the project repo to be released'
+      project_version:
+        description: 'Version number to be released e.g 10.0.3'
+      fancy:
+        description: 'If fancy versioning is required set it "True"'
+        default: 'False'
+      fancy_version_no:
```
hanefi, line 11: Why do we need both `fancy` and `fancy_version_no`? I do not follow the logic here.

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,60 @@
...
+        description: 'If fancy versioning is True, then it should be set e.g 1'
+        default: '1'
+      microsoft_email:
+        description: 'Email to be written on changelogs'
+        default: 'gindibay@microsoft.com'
+      name:
+        description: 'Name to be written on changelogs'
+        default: 'Gurkan Indibay'
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
```
hanefi:
```suggestion
```

I think this is a well known concept and we don't need this explanation here.

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,60 @@
...
+jobs:
+  build:
+    # The type of runner that the job will run on
```
hanefi:
```suggestion
```

`runs-on: ubuntu-latest` is really clear and it does not need any comments IMO

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,60 @@
...
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
```
hanefi:
```suggestion
```

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,60 @@
...
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      - name: Clone Tools branch
+        run: git clone --branch tools-transition https://github.com/citusdata/tools.git
+
+      - name: install requirements
```
hanefi:
```suggestion
      - name: Install Python requirements
```

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,60 @@
...
+        run: python -m pip install -r tools/python/requirements.txt
+
+      - name: Execute Package Properties Update
```
hanefi:
```suggestion
      - name: Update Package Properties
```

Does this make sense?

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,60 @@
...
+        run: |
+              git config --global user.email "${MICROSOFT_EMAIL}"
+              git config --global user.name "${NAME}"
+              chmod +x ./tools/bash/update_package_properties.sh && ./tools/bash/update_package_properties.sh
```
hanefi: 2 questions:

1. Why do we need `chmod +x` ? Shouldn't we set this file mode on the tools repo?
2. Why is this script not in a new directory and not in `package_automation`?

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,60 @@
...
+  FANCY_VERSION_NO: ${{ github.event.inputs.fancy_version_no }}
+  MICROSOFT_EMAIL: ${{ github.event.inputs.microsoft_email }}
+  NAME: ${{ github.event.inputs.name }}
+
+# Controls when the action will run.
+on:
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+    inputs:
```
JelteF, line 8: Some general question about this PR.
1. Do I understand correctly that we will manually trigger a build of this workflow (by clicking in the interface) to create a PR for a new package version?
2. What is the repo layout of that we want in the future? Still one branch per project? If so, then I think having most of these values be configurable is not very useful. They can be hardcoded in the workflow since each workflow will have a different template.

gurkanindibay: Answers
1. Yes it is true
2. Yes I don't think to change it unless there is a better solution for packaging. Our pipelines are not configurable. Only getting parameters from users. For example in this pipeline project_name, tag_name, project_version, fancy and fancy_version, microsoft_email and name is being requested from user. Git hub token is a secret value so it is being stored on and referenced from Repository Secret Store. In summary, we can not hardcode parameters because most of the parameters are being requested from user and the other one is already hardcoded but hardoced on Github Repository Secret Store

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed

.github/workflows/update_package_properties.yml
gurkanindibay: Fancy parameter is needed for whether fancy parameter required or not. If required, fancy-version_no parameter is used as fancy versioning extension

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed

.github/workflows/update_package_properties.yml
gurkanindibay: I'm not sure but when checking out manually as here, there may be some permission problems. I had the problem when I want to execute "./tools/bash/update_package_properties.sh" command so that I needed to add this line.
I tried to push the permissions but I couldn't find a way to do it

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,52 @@
...
+  PRJ_NAME: ${{ github.event.inputs.project_name }}
+  TAG_NAME: ${{ github.event.inputs.tag_name }}
+  PRJ_VER: ${{ github.event.inputs.project_version }}
+  FANCY: ${{ github.event.inputs.fancy }}
+  FANCY_VERSION_NO: ${{ github.event.inputs.fancy_version_no }}
+  MICROSOFT_EMAIL: ${{ github.event.inputs.microsoft_email }}
+  NAME: ${{ github.event.inputs.name }}
+
+on:
+
```
hanefi:
```suggestion
```

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,52 @@
...
+        description: 'If fancy versioning is True, then it should be set e.g 1'
+        default: '1'
+      microsoft_email:
+        description: 'Email to be written on changelogs'
+        default: 'gindibay@microsoft.com'
+      name:
+        description: 'Name to be written on changelogs'
+        default: 'Gurkan Indibay'
+jobs:
+  build:
```
hanefi: I think we can rename this job name. We do not actually build anything. You can also supply a better name so we have a better name in the GitHub action run outputs.

See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idname

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,52 @@
...
+  FANCY: ${{ github.event.inputs.fancy }}
+  FANCY_VERSION_NO: ${{ github.event.inputs.fancy_version_no }}
+  MICROSOFT_EMAIL: ${{ github.event.inputs.microsoft_email }}
+  NAME: ${{ github.event.inputs.name }}
+
+on:
+
+  workflow_dispatch:
+    inputs:
+      project_name:
```
hanefi: Since you wish to merge these changes into `all-citus` branch, won't the project name always be Citus? Do you wish to apply similar changes in other `all-X` branches as well?

I suggest we keep different workflow definitions for each project, and hardcode such parameters where possible. 
(i) This way we will have one less thing to worry about on a release since we will not need to swap between different projects
(ii) This way we will not lose our ability to use different versions of tools repository for different projects, and thus we will not break package creation for the less frequently updated projects such as topn and hll.

hanefi (commented):
We do not really need single or double quotation marks except for the ones in last 3 lines.

.github/workflows/update_package_properties.yml
hanefi: This is actually an example of what @JelteF  asked earlier

> What is the repo layout of that we want in the future? Still one branch per project? If so, then I think having most of these values be configurable is not very useful. They can be hardcoded in the workflow since each workflow will have a different template.

gurkanindibay: Removed projet_name parameter and hardocoded project_name as "citus"

.github/workflows/update_package_properties.yml
gurkanindibay: Changed name into update_package_properties

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed as you suggested

.github/workflows/update_package_properties.yml
gurkanindibay: @hanefi suggested me to hardcode project_name since project_name is not dynamic on this branch

hanefi (approved):
Please squash your changes into a single commit before merging this.

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,52 @@
...
+  FANCY_VERSION_NO: ${{ github.event.inputs.fancy_version_no }}
+  MICROSOFT_EMAIL: ${{ github.event.inputs.microsoft_email }}
+  NAME: ${{ github.event.inputs.name }}
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: 'The tag name on the project repo to be released'
+      project_version:
```
JelteF: I think the project version should be derived from the tag name.

.github/workflows/update_package_properties.yml
JelteF: I think `fancy` should also be hardcoded, because that should stay the same for the same project.

gurkanindibay: @hanefi what do you think of hardcoding Fancy parameter? I understood from your comments that fancy versioning is not needed all the times. It is useful when there is a need for adding a new release without changing version number

.github/workflows/update_package_properties.yml
gurkanindibay: @hanefi what is your opinion about it? Is there any exceptional case that breaks the relation between version and tag? If not I can remove the parameter version and correlate that varible with tag by removing 'v' prefix.

hanefi: I think it makes sense to do this. The package version and the tag name will always be the same (except for the `v` prefix)

.github/workflows/update_package_properties.yml
hanefi: It makes sense to have a hardcoded `fancy` value for projects. And a `fancy_version` parameter.

As far as I recall, all our extensions use fancy versioning today, and it is unlikely that we will change that in the near future.

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,52 @@
...
+      PLATFORM: ${{ matrix.platform }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      #      - name: install dependencies
+      #        run: sudo apt install libcurl4-openssl-dev libssl-dev
+      #
+      #      - name: Define git credentails
+      #        run: git config --global user.email "gindibay@microsoft.com"&& git config --global user.name "Gurkan Indibay"
```
hanefi: stale comments

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,52 @@
...
+
+  workflow_dispatch:
+
+jobs:
+  build_package:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: [ el/7, el/8, ol/7, debian/stretch, debian/buster, ubuntu/xenial, ubuntu/bionic,ubuntu/focal ]
```
hanefi: inconsistent spacing.

It'd be easier to read if we have one item per line, I think.

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,52 @@
...
+#        platform: [ debian/stretch ]
```
hanefi: stale comment

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: sudo apt install libcurl4-openssl-dev libssl-dev
+
+      - name: Install python requirements
+        run: python -m pip install -r tools/packaging_automation/requirements.txt
+
+      - name: Build Packages
+        run: |
+          python -m  tools.packaging_automation.citus_package --gh_token "${{ secrets.GH_TOKEN }}" --platform "${{ matrix.platform }}" \
+          --build_type 'release' --secret_key "${{ secrets.PACKAGING_SECRET_KEY }}" --passphrase "${{ secrets.PACKAGING_PASSPHRASE }}" \
+          --output_dir "$(pwd)/packages/" --input_files_dir "$(pwd)" --output_validation "False"
```
hanefi: I have trouble reading this. How about splitting every parameter in a new line?

```suggestion
          python -m tools.packaging_automation.citus_package \
                 --gh_token "${{ secrets.GH_TOKEN }}" \
                 --platform "${{ matrix.platform }}" \
                 --build_type 'release' \
                 --secret_key "${{ secrets.PACKAGING_SECRET_KEY }}" \
                 --passphrase "${{ secrets.PACKAGING_PASSPHRASE }}" \
                 --output_dir "$(pwd)/packages/" \
                 --input_files_dir "$(pwd)" \
                 --output_validation "False"
```

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Publish Packages
+        run: |
+          python -m  tools.packaging_automation.upload_to_package_cloud --platform ${{ matrix.platform }} --package_cloud_api_token ${{ secrets.PACKAGE_CLOUD_API_TOKEN }} \
+          --repository_name "sample" --output_file_path "$(pwd)/packages"
```
hanefi: I do not see a secret for `PACKAGE_CLOUD_API_TOKEN` in the env section. How does this actually work?

Also this is hard to read for me.

```suggestion
          python -m  tools.packaging_automation.upload_to_package_cloud \
                 --platform ${{ matrix.platform }} \
                 --package_cloud_api_token ${{ secrets.PACKAGE_CLOUD_API_TOKEN }} \
                 --repository_name "sample" \
                 --output_file_path "$(pwd)/packages"
```

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,52 @@
+name: Build Package
+
+env:
```
hanefi, line 3: Do we need `PACKAGE_CLOUD_API_TOKEN` here?

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,50 @@
...
+
+    steps:
+      - name: Checkout repository
+      - uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: sudo apt install libcurl4-openssl-dev libssl-dev
+
+      - name: Clone Tools branch
+        run: git clone --branch tools-transition https://github.com/citusdata/tools.git
```
hanefi: We can add `--depth=1` for a faster operation

Also how about using `develop` for now, and starting to use tags after we are done with the whole transition?

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Install Python requirements
+        run: python -m pip install -r tools/python/requirements.txt
+
+      - name: Update Package Properties
```
hanefi: Casing

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      #      - name: install dependencies
+      #        run: sudo apt install libcurl4-openssl-dev libssl-dev
+      #
+      #      - name: Define git credentails
+      #        run: git config --global user.email "gindibay@microsoft.com"&& git config --global user.name "Gurkan Indibay"
+
+      - name: Clone tools branch
+        run: git clone --depth=1 --branch tools-transition-update-docker https://github.com/citusdata/tools.git tools
```
hanefi: It would be nicer to create a new tag on the tools repo, and checkout the tag. We can clone develop for now, but we need to start using tags at some poing

.github/workflows/update_package_properties.yml
hanefi: See https://github.com/citusdata/tools/pull/102 for my fix. After that one is merged and we switch back to checking out `develop` branch of tools, we will no longer need setting execution permissions here.

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,50 @@
...
+        run: |
+          git config --global user.email "${MICROSOFT_EMAIL}"
+          git config --global user.name "${NAME}"
+          chmod +x ./tools/bash/update_package_properties.sh && ./tools/bash/update_package_properties.sh
```
hanefi: We will not need the `chmod` part after this one is merged: https://github.com/citusdata/tools/pull/102

.github/workflows/build-package.yml
gurkanindibay: Fixed

.github/workflows/update_package_properties.yml
gurkanindibay: removed fancy parameter and using fancy_version parameter as a fancy switch. If it is greater than 1 than fancy is true.

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed

.github/workflows/build-package.yml
gurkanindibay: Fixed

.github/workflows/build-package.yml
gurkanindibay: Fixed

.github/workflows/build-package.yml
gurkanindibay: Fixed

.github/workflows/build-package.yml
gurkanindibay: Fixed

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,60 @@
...
+      fail-fast: false
+      matrix:
+        platform: [ el/7,
+                    el/8,
+                    ol/7,
+                    debian/stretch,
+                    debian/buster,
+                    ubuntu/xenial,
+                    ubuntu/bionic,
+                    ubuntu/focal ]
```
hanefi:
```suggestion
        platform:
          - el/7,
          - el/8,
          - ol/7,
          - debian/stretch,
          - debian/buster,
          - ubuntu/xenial,
          - ubuntu/bionic,
          - ubuntu/focal
```

IMO arrays are easier to read when we drop `[]` and use yaml features for lists instead.

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,60 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Clone tools branch
+        run: git clone --depth=1 --branch develop https://github.com/citusdata/tools.git tools
```
hanefi: I suggest you bump a new version on tools repo and use that tag so that it will not break as easily in the future.

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: sudo apt install libcurl4-openssl-dev libssl-dev
+
+      - name: Clone tools branch
+        run: git clone --branch update_properties_fixes https://github.com/citusdata/tools.git
```
hanefi: Do we want to keep using this custom branch?

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Install python requirements
+        run: python -m pip install -r tools/packaging_automation/requirements.txt
+
+      - name: Update package Properties
+        run: |
+              python -m packaging_automation.update_package_properties --gh_token=${{ secrets.GH_TOKEN }} \
+              --prj_name "${PRJ_NAME}" --tag_name ${{ github.event.inputs.tag_name }} \
+              --email ${{ github.event.inputs.microsoft_email }} --name ${{ github.event.inputs.name }} --pipeline \
+              --exec_path "$(pwd)"
```
hanefi: Please have one argument per line.

hanefi (changes requested):

.github/workflows/update_package_properties.yml
gurkanindibay: Removed permission grant command as you suggested

.github/workflows/build-package.yml
gurkanindibay: Created v0.8.0 tag and used this tag in pipelines as tools source

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed using 0.8.0

.github/workflows/build-package.yml
gurkanindibay: Fixed

.github/workflows/build-package.yml
gurkanindibay: Fixed

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed using tag

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,44 @@
...
+        description: 'If fancy versioning is needed, set value greater than 1 e.g 2'
+        default: '1'
+      microsoft_email:
+        description: 'Email to be written on changelogs'
+        default: 'gindibay@microsoft.com'
+      name:
+        description: 'Name to be written on changelogs'
+        default: 'Gurkan Indibay'
+jobs:
+  update_package_properties:
```
hanefi, line 21:
```suggestion
  update_package_properties:
    name: Update Package Properties
```

I left a comment https://github.com/citusdata/packaging/pull/645#discussion_r611405107 with 2 parts. You addressed the first item there, and missed the second one

.github/workflows/update_package_properties.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Install dependencies
+        run: sudo apt install libcurl4-openssl-dev libssl-dev
+
+      - name: Clone tools branch
+        run: git clone -b v0.8.0 --depth=1  https://github.com/citusdata/tools.git tools
+
+      - name: Install python requirements
+        run: python -m pip install -r tools/packaging_automation/requirements.txt
+
+      - name: Update package Properties
```
hanefi: casing

hanefi (changes requested):
Overall looks good. However needs some more polishing.

There are several inconsistencies across different GH pipelines.
1. Supply human readable `name:` fields when it makes sense so that it is clear and easier to understand for the end user that will trigger the workflow.
2. Have consistent casing. I do not really have a strict preference, but I'd like to see some consistency across the board.
3. Have consistent formatting for string values. Do you wish to use single quotes, double quotes or no quotes? Please pick one and be consistent. If possible use some linters to enforce these rules on your editors for the future.

I used https://prettier.io/ cli with default configs and it decided to do the following changes:

```diff
diff --git a/.github/workflows/update_package_properties.yml b/.github/workflows/update_package_properties.yml
index 6ff9935..4f0ab91 100644
--- a/.github/workflows/update_package_properties.yml
+++ b/.github/workflows/update_package_properties.yml
@@ -6,16 +6,16 @@ on:
   workflow_dispatch:
     inputs:
       tag_name:
-        description: 'The tag name on the project repo to be released'
+        description: "The tag name on the project repo to be released"
       fancy_version_no:
-        description: 'If fancy versioning is needed, set value greater than 1 e.g 2'
-        default: '1'
+        description: "If fancy versioning is needed, set value greater than 1 e.g 2"
+        default: "1"
       microsoft_email:
-        description: 'Email to be written on changelogs'
-        default: 'gindibay@microsoft.com'
+        description: "Email to be written on changelogs"
+        default: "gindibay@microsoft.com"
       name:
-        description: 'Name to be written on changelogs'
-        default: 'Gurkan Indibay'
+        description: "Name to be written on changelogs"
+        default: "Gurkan Indibay"
 jobs:
   update_package_properties:
     runs-on: ubuntu-latest
@@ -35,10 +35,10 @@ jobs:

       - name: Update package Properties
         run: |
-              python -m packaging_automation.update_package_properties \
-              --gh_token=${{ secrets.GH_TOKEN }} \
-              --prj_name "${PRJ_NAME}" \
-              --tag_name ${{ github.event.inputs.tag_name }} \
-              --email ${{ github.event.inputs.microsoft_email }} \
-              --name ${{ github.event.inputs.name }} --pipeline \
-              --exec_path "$(pwd)"
+          python -m packaging_automation.update_package_properties \
+          --gh_token=${{ secrets.GH_TOKEN }} \
+          --prj_name "${PRJ_NAME}" \
+          --tag_name ${{ github.event.inputs.tag_name }} \
+          --email ${{ github.event.inputs.microsoft_email }} \
+          --name ${{ github.event.inputs.name }} --pipeline \
+          --exec_path "$(pwd)"
```

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed

gurkanindibay: Fixed

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed

.github/workflows/update_package_properties.yml
gurkanindibay: Fixed

.github/workflows/update_package_properties.yml
hanefi: How is this fixed exactly? I do not see any field for `jobs.update_package_properties.name` here.

hanefi (changes requested):

.github/workflows/update_package_properties.yml
gurkanindibay: I hadn't realized the job name field until you clarify with the comment above. I didn't get used to the structure of comments half is here half is up somewhere.
By the way, when you solve the comments I can track the comments more easily.

gurkanindibay: Fixed as you suggested

hanefi (approved):

PR #655, opened 2021-05-24
Postgres repo install problem

### LINKED ISSUE #652: rpm.sh might fail to find postgresql package even if it exists
https://github.com/citusdata/packaging/blob/gh-pages/enterprise/rpm.sh#L47..L48

Here we execute `yum info` to check if postgresql package is available on pgdg repo.
`yum info` might sometimes ask for user input to install GPG key for pgdg repo.
However, since we redirect all the output to `/dev/null`, users might see such false errors:

>PGDG repositories don't have postgresql13-server package for your operating system
Cannot install Citus, exiting.

This is because, `yum info` expects user to provide 'y' to install GPG key, but we don't already show that to users.

To reproduce, simply run `rpm.sh` on centos7.9 docker image.

gurkanindibay:
Fixes #652

### DIFF
--- a/community-nightlies/rpm.sh
+++ b/community-nightlies/rpm.sh
@@ -44,7 +44,7 @@ pgdg_check ()
 
     yum install -d0 -e0 -y "${repo_url}"
 
-    if ! yum info -q postgresql13-server &> /dev/null; then
+    if ! yum info -q -y postgresql13-server &> /dev/null; then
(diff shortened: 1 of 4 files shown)


onurctirtir (approved):

hanefi (approved):
👍

gurkanindibay:
### Before
image
### After
image

PR #657, opened 2021-07-08
Bump to citus 9.4.5

gurkanindibay:


.travis.yml
```diff
@@ -29,7 +29,7 @@ before_install:
   - sudo make -C tools install
   - export PACKAGING_SECRET_KEY="$(openssl aes-256-cbc -K $encrypted_a39e3f1ee8ba_key -iv $encrypted_a39e3f1ee8ba_iv -in signing_key.asc.enc -d | base64)"
 install: true
-script: build_new_release && build_new_nightly
```
hanefi, line 32: I'd like it more if we comment the second part instead of deleting it. Preferably with a comment so that we do not forget to revert it after the next releases.

citus.spec
```diff
@@ -79,6 +79,9 @@ echo %{pginstdir}/share/extension/%{sname}.control >> installation_files.list
 %doc %{pginstdir}/doc/extension/README-%{sname}.md
 
 %changelog
+* Thu Jul 08 2021 - Gurkan <gindibay@microsoft.com> 9.4.5.citus-1
+- Official 9.4.5 release of Citus
+
 * Thu Mar 18 2021 - Gurkan Indibay <gindibay@microsoft.com> 10.0.3.citus-1
```
hanefi, line 85: We are not always consistent with our changelog contents when we release a backport.

Does it really make sense to have changelog items for 10.x on a 9.y package? I do not think so.

A solution is to create this branch right after the commit that created 9.4.4 packages. This way, it will not have any entries for 10.x. It will still contain 9.5.1 which is a bummer.

All in all, I am fine with this file now, and this is only because it will take so much effort and it is not worth fixing this IMO.

debian/changelog
```diff
@@ -1,3 +1,17 @@
...
+  
+  * Fixes a bug that causes pruning incorrect shard of a range distributed table
+  
+  * Fixes a bug that might cause self-deadlocks when COPY used in TX block
+  
+  * Fixes an issue that could cause citus_finish_pg_upgrade to fail
+  
+ -- Gurkan <gindibay@microsoft.com>  Thu, 08 Jul 2021 10:15:23 +0000  
+ 
 citus (10.0.3.citus-1) stable; urgency=low
```
hanefi, line 15: See my comment on `citus.spec` file where I shared my take on having 10.x entries in 9.y package changelogs.

hanefi (commented):
I really like the way you split your work in meaningful and small commits.

citus.spec
gurkanindibay: There may be some all-citus-9, all-citus-10 and we can continue on this branch. After these releases we can plan such structure

.travis.yml
gurkanindibay: Fixed

debian/changelog
gurkanindibay: I put a comment on that comment

hanefi (approved):

hanefi:
I should note here that we dropped support for el6 and ol6 since `9.4.4` release. So these distros will not get the packages for `9.4.5`. We had a discussion offline and this makes sense.

If we were to create the branches from the tip of `9.4.4`, this might not be the case. See more discussion [here](https://github.com/citusdata/packaging/pull/657#discussion_r666166058)

PR #659, opened 2021-07-09
Bump to citus 9.5.6

gurkanindibay:


citus.spec
```diff
@@ -7,11 +7,11 @@ Summary:	PostgreSQL-based distributed RDBMS
 Name:		%{sname}%{?pkginfix}_%{pgmajorversion}
 Provides:	%{sname}_%{pgmajorversion}
 Conflicts:	%{sname}_%{pgmajorversion}
-Version:	9.4.5.citus
+Version:	9.5.6.citus
 Release:	1%{dist}
 License:	AGPLv3
 Group:		Applications/Databases
-Source0:	https://github.com/citusdata/citus/archive/v10.0.3.tar.gz
```
hanefi, line 14: I do not like that we did not really update this field when releasing 9.4.5 packages. We need to check if those packages are ok.

See https://github.com/citusdata/packaging/pull/657/files#diff-96b326beae79b577177ee0e7fe26e169c66156d17bc9fa0b58ca4536fc7010e2L14

pkgvars
```diff
@@ -1,6 +1,6 @@
 pkgname=citus
 pkgdesc='Citus (Open-Source)'
-pkglatest=9.4.5.citus-1
-releasepg=11,12
-nightlypg=11,12
+pkglatest=9.5.6.citus-1
+releasepg=11,12,13
+nightlypg=12,13
```
hanefi, line 5: We already disabled the nightly builds. Does it really make sense to update the supported PG versions here?

gurkanindibay:
I closed this one since we needed to repackage 9.4.5 after opening this PR

PR #671, opened 2021-07-09
Bump to citus 9.5.6

gurkanindibay:


pkgvars
```diff
@@ -1,6 +1,6 @@
 pkgname=citus
 pkgdesc='Citus (Open-Source)'
-pkglatest=9.4.5.citus-2
-releasepg=11,12
-nightlypg=11,12
+pkglatest=9.5.6.citus-1
+releasepg=11,12,13
+nightlypg=12,13
```
hanefi, line 5: Updating nightly PG versions do not have any effect. However they are harmless so we can let this one pass, because it makes sense for the script to set it to `12,13` here.

hanefi (approved):
:shipit:

PR #683, opened 2021-07-16
Bump to citus_enterprise 10.0.4

gurkanindibay:


pkgvars
```diff
@@ -1,6 +1,6 @@
 pkgname=citus-enterprise
 pkgdesc='Citus Enterprise'
-pkglatest=9.5.6.citus-1
+pkglatest=10.0.4.citus-1
 releasepg=11,12,13
 nightlypg=11,12,13
```
hanefi, line 5:
```suggestion
nightlypg=12,13
```

`enterprise-master` no longer supports pg11. However I do not think that we create nightly packages for enterprise so this may not have any effect at all.

hanefi (approved):
Left one comment that is completely optional.

PR #691, opened 2021-08-06
Fix ModuleNotFoundError for update_package_properties

gurkanindibay:
Fixes #690 
Blocked on https://github.com/citusdata/tools/pull/150

Additionally fixes git commit problem while executing update_package_properties in pipeline since git username and email is required for commit
To be able to successfully execute this pipeline tools issue 149 should be first reviewed and closed. After merging it, tagging tools code and using the tagged code here will be required. After finishing both of the PRs, at the end I will change the tools tag reference here

PR #706, opened 2021-08-10
Adds update_package_properties and build_packages pipelines into all-enterprise

gurkanindibay:


hanefi:
I think we should complete all open PRs for `all-citus` branch and then open a PR for `all-enterprise`.

For example https://github.com/citusdata/packaging/pull/691 is still a work in progress, and I would want to see similar changes in this branch as well.

Please wait until the development for citus community packages are complete before updating this PR

gurkanindibay:
> I think we should complete all open PRs for `all-citus` branch and then open a PR for `all-enterprise`.
> 
> For example #691 is still a work in progress, and I would want to see similar changes in this branch as well.
> 
> Please wait until the development for citus community packages are complete before updating this PR

I think they're independent you can continue on

hanefi (approved):
I compared the contents of the changed files with those in `all-citus` and it makes sense the way it is.

PR #709, opened 2021-08-10
Adds update_package_properties and build_packages pipelines into pg-auto-failover

gurkanindibay:


hanefi:
@gurkanindibay can you check the failed Ubuntu Xenial tests? Let me remind you that it is EOL'd and we can and maybe we should remove it from our build matrix.

gurkanindibay:
> @gurkanindibay can you check the failed Ubuntu Xenial tests? Let me remind you that it is EOL'd and we can and maybe we should remove it from our build matrix.

I checked the failed test and it's kinda weird. it's about python version that is ubuntu xenial image using. But it does not fail on Travis

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,62 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - el/7
+          - el/8
+          - ol/7
+          - debian/stretch
+          - debian/buster
+          - ubuntu/xenial
```
hanefi: I think we should remove this line

```suggestion
```

hanefi: similarly we can update `.travis.yml`, preferably in a separate PR.

gurkanindibay: Asking @DimCitus might be useful

hanefi (approved):

PR #715, opened 2021-08-11
Build and publish pipeline for pgxn

gurkanindibay:


.github/workflows/build_package.yml
```diff
@@ -0,0 +1,28 @@
...
+          PGXN_USERNAME: ${{ secrets.PGXN_USER_NAME }}
+          PGXN_PASSWORD: ${{ secrets.PGXN_PASSWORD }}
+        run: |
+          CURRENT_BRANCH="${GITHUB_REF##*/}"
+          if [ "${CURRENT_BRANCH}" = "${MAIN_BRANCH}" ]
+          then
+            pgxn-release
+          else
+            echo "Distro will not be released since current branch '${CURRENT_BRANCH}' is different from main branch '${MAIN_BRANCH}'"
+          fi
```
hanefi, line 28:
```suggestion
        if: ${{ github.ref == ${MAIN_BRANCH} }}
        run: pgxn-release
```

I haven't checked if I need some more quotation marks there. Please double check if this makes sense

.github/workflows/build_package.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  release:
+    name: Release on GitHub and PGXN
+    runs-on: ubuntu-latest
+    container: pgxn/pgxn-tools
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Bundle the Release
+        id: bundle
```
hanefi: Why do we need to supply an `id` field here? It appears unused.

gurkanindibay: Fixed

.github/workflows/build_package.yml
gurkanindibay: It will not work since github.ref is /refs/HEAD/{branch_name}.

gurkanindibay: I want to see the detailed message when it is not pushed as well

hanefi (approved):

PR #722, opened 2021-08-17
Bump to citus_enterprise 9.5.7

gurkanindibay:


pkgvars
```diff
@@ -1,7 +1,7 @@
 pkgname=citus-enterprise
 pkgdesc='Citus Enterprise'
-pkglatest=9.4.6.citus-1
-releasepg=11,12
-nightlypg=11,12
+pkglatest=9.5.7.citus-1
+releasepg=11,12,13
+nightlypg=11,12,13
```
hanefi, line 5: We can not generate PG11 nightly images for our projects. However this is not really urgent to fix here.

We can fix that when releasing `10.1.2` packages, that should happen soon.

hanefi (approved):
one minor comment that we can address later. No urgency here

PR #729, opened 2021-08-18
Bump pgxn to version 10.1.2

gurkanindibay:


pkgvars
```diff
@@ -1,4 +1,4 @@
 pkgname=citus
 pkgdesc='Citus (Open-Source)'
-pkglatest=10.1.1
+pkglatest=10.1.2
 releasepg=11,12,13
```
hanefi:
```suggestion
releasepg=12,13
```

hanefi: We may need to update some other PG versions in `META.json` as well. E.g. https://github.com/citusdata/packaging/blob/92120972f518c2f3540110f64e7a56bda34d6cdc/META.json#L19

hanefi: @gurkanindibay can you update PG version in `META.json` as well?

It appears that you missed my second comment here.

hanefi (approved):

PR #731, opened 2021-08-18
Fix for error in fetch_and_build_deb unbound variable error

### LINKED ISSUE #730: error while building ubuntu packages
There was an error in fetch_and_build_deb causes build to fail. Error is  
 ```
/scripts/fetch_and_build_deb: line 185: file: unbound variable 
```
Build with the errror is below
https://github.com/citusdata/tools/runs/3360692065?check_suite_focus=true

### PR DESCRIPTION
**Pull request by gurkanindibay**

### CODE REVIEW

File: scripts/fetch_and_build_deb [hunk 1]
```diff
@@ -176,14 +176,13 @@ mjobs="$(expr $procs + 1)"
...
 
 ## Rename all *.ddeb files to *.deb since we upload debug packages with .deb suffix
-for filename in ../*.ddeb
-do
+for file in ../*.ddeb; do
   # .ddeb suffix is automatically chosen when naming debug packages for Ubuntu
   # releases. For this reason, we rename all *.ddeb files to *.deb since we upload
   # debug packages with .deb suffix.
-  if [ ${filename} != "../*.ddeb" ]; then
+  if [ "${file}" != "../*.ddeb" ]; then
```
hanefi: I still do not get why we need this checks here. Do you know how the `file` variable can store the string `../*.ddeb` ?

onurctirtir: https://stackoverflow.com/a/6364244/13370386

gurkanindibay: if ddeb file does not exist, filename becomes ../*.ddeb

hanefi:
```suggestion
  if [ -e "${file}" ]; then
```

I think this makes it easier to understand what is going on.

hanefi (approved):
I have a suggestion for the template that needs to be copied to all bash scripts.

scripts/fetch_and_build_deb
gurkanindibay: Fixed

PR #738, opened 2021-08-25
Adds postgres support matrix support for debian based systems

gurkanindibay:
Build parametrization according to postgres version is handled in docker images for debian based systems since there is only one image for them. Versions should be added into debian/pgversions to be built. Before this PR these parameters are being taken from pkgvars. With this PR , I got these paramaters postgres-versions.txt file.

### DIFF
--- a/dockerfiles/ubuntu-xenial-all/scripts/fetch_and_build_deb
+++ b/dockerfiles/ubuntu-xenial-all/scripts/fetch_and_build_deb
@@ -49,19 +49,26 @@ fi
 # populate variables from packaging metadata file
 # shellcheck source=/dev/null
 source /buildfiles/pkgvars
-source /buildfiles/supported-postgres.txt
-# set default values for certain packaging variables
+# Read PostgreSQL versions from file.
+# supported-postgres file is derived file from postgres-matrix.yml file by citus_package.
+# If file does not exist, skip populating file
+[[ -f "/buildfiles/supported-postgres" ]] && source /buildfiles/supported-postgres
+
+# Fetch pkgname, hubproj, nightlyref, versioning from pkgvars file
 declare pkglatest # to make shellcheck happy
 pkgname="${deb_pkgname:-${pkgname}}"
 hubproj="${hubproj:-${pkgname}}"
 nightlyref="${nightlyref:-master}"
-releasepg="${release_versions:-${releasepg}}"
-nightlypg="${nightly_versions:-${nightlypg}}"
 versioning="${versioning:-simple}"
+# Fetch pg release and nightly versions from supported-postgres file which is originated from postgres-matrix.yml file
+# If this file is not found, releasepg and nightlypg parameters from pkgvars are used for defining pg versions
+pg_release_versions="${release_versions:-${releasepg}}"
+pg_nightly_versions="${nightly_versions:-${nightlypg}}"
+
 
 echo "Postgres versions:"
-echo "Release Versions:${releasepg}"
-echo "Nightly Versions:${nightlypg}"
+echo "Release Versions: ${pg_release_versions}"
+echo "Nightly Versions: ${pg_nightly_versions}"
 
 if [ -z "${pkglatest}" ]; then
     echo "$0: pkgvars file must specify a value for pkglatest" >&2
@@ -100,7 +107,7 @@ case "${1}" in
             exit $failure
         fi
 
-        echo "${releasepg}" | tr ',' '\n' > "${builddir}/debian/pgversions"
+        echo "${pg_release_versions}" | tr ',' '\n' > "${builddir}/debian/pgversions"
         ;;
     *)
         if [ "${1}" == 'nightly' ]; then
@@ -128,7 +135,7 @@ case "${1}" in
         packageversion="${nextversion}.citus~${packagesuffix}"
         export CONF_EXTRA_VERSION="+${packagesuffix}"
 
-        echo "${nightlypg}" | tr ',' '\n' > "${builddir}/debian/pgversions"
+        echo "${pg_nightly_versions}" | tr ',' '\n' > "${builddir}/debian/pgversions"
         ;;
 esac
 
--- a/dockerfiles/ubuntu-xenial-all/scripts/fetch_and_build_deb
+++ b/dockerfiles/ubuntu-xenial-all/scripts/fetch_and_build_deb
@@ -55,9 +55,10 @@ declare pkglatest # to make shellcheck happy
 pkgname="${deb_pkgname:-${pkgname}}"
 hubproj="${hubproj:-${pkgname}}"
 nightlyref="${nightlyref:-master}"
-releasepg="${releasepg:-11,12,13}"
-nightlypg="${nightlypg:-${releasepg}}"
+
 versioning="${versioning:-simple}"
+supported_postgres_versions="${postgres_versions:-11,12,13}"
+nightlypg="${supported_postgres_versions}"
 
 if [ -z "${pkglatest}" ]; then
     echo "$0: pkgvars file must specify a value for pkglatest" >&2
@@ -96,7 +97,7 @@ case "${1}" in
             exit $failure
         fi
 
-        echo "${releasepg}" | tr ',' '\n' > "${builddir}/debian/pgversions"
+        echo "${supported_postgres_versions}" | tr ',' '\n' > "${builddir}/debian/pgversions"
         ;;
     *)
         if [ "${1}" == 'nightly' ]; then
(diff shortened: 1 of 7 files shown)


dockerfiles/ubuntu-xenial-all/scripts/fetch_and_build_deb
```diff
@@ -49,16 +49,20 @@ fi
 # populate variables from packaging metadata file
 # shellcheck source=/dev/null
 source /buildfiles/pkgvars
-
+source /buildfiles/supported-postgres.txt
```
hanefi: I think we should remove all files for ubuntu/xenial from this repo. Can you do that in a separate PR and update this one accordingly?

scripts/fetch_and_build_deb
```diff
@@ -49,16 +49,20 @@ fi
 # populate variables from packaging metadata file
 # shellcheck source=/dev/null
 source /buildfiles/pkgvars
```
hanefi, line 51: Should we remove this line as well, if we want to stop using `pkgvars`?

scripts/fetch_and_build_deb
```diff
@@ -49,16 +49,20 @@ fi
...
-
+source /buildfiles/supported-postgres.txt
 # set default values for certain packaging variables
 declare pkglatest # to make shellcheck happy
```
hanefi, line 58: What will happen to `pkglatest` when we remove `pkgvars`?

scripts/fetch_and_build_deb
gurkanindibay: getting pkglatest still from pkgvars

scripts/fetch_and_build_deb
gurkanindibay: We will not move pkgvars. We still need to get parameters other than postgres version from pkgvars

dockerfiles/ubuntu-xenial-all/scripts/fetch_and_build_deb
gurkanindibay: https://github.com/citusdata/packaging/issues/744

scripts/fetch_and_build_deb
```diff
@@ -49,16 +49,20 @@ fi
 # populate variables from packaging metadata file
 # shellcheck source=/dev/null
 source /buildfiles/pkgvars
-
+source /buildfiles/supported-postgres.txt
```
hanefi: I have 2 concerns here:

1. `source` return value is non-zero if the file is not found. This will make the script fail as we have `set -e` option in the file.

```bash
❯ source no-file-with-this-name
source: no such file or directory: no-file-with-this-name
❯ echo $?
127
```

2. We do not need the `.txt` extension. This is Linux territory and looks weird to have extensions.

scripts/fetch_and_build_deb
```diff
@@ -49,16 +49,20 @@ fi
...
 # set default values for certain packaging variables
 declare pkglatest # to make shellcheck happy
 pkgname="${deb_pkgname:-${pkgname}}"
 hubproj="${hubproj:-${pkgname}}"
 nightlyref="${nightlyref:-master}"
-releasepg="${releasepg:-11,12,13}"
-nightlypg="${nightlypg:-${releasepg}}"
+releasepg="${release_versions:-${releasepg}}"
```
hanefi: I think your intention is not so easy to read here. You need to explain this better.

Do you want the `release_versions` from to overwrite the value supplied in `releasepg`? Can you do that in a cleaner fashion?

scripts/fetch_and_build_deb
```diff
@@ -49,16 +49,20 @@ fi
...
+nightlypg="${nightly_versions:-${nightlypg}}"
 versioning="${versioning:-simple}"
 
+echo "Postgres versions:"
+echo "Release Versions:${releasepg}"
+echo "Nightly Versions:${nightlypg}"
```
hanefi:
```suggestion
echo "Release Versions :${releasepg}"
echo "Nightly Versions :${nightlypg}"
```

hanefi (changes requested):

scripts/fetch_and_build_deb
gurkanindibay: 1. I see. my intention is not to fail the script in case of file's nonexistence. So I think I need to handle the error here.

scripts/fetch_and_build_deb
gurkanindibay: My intention is to use pkgvars releasepg values if file not found

scripts/fetch_and_build_deb
gurkanindibay: Fixed

scripts/fetch_and_build_deb
gurkanindibay: Changed variable names and added some comments

gurkanindibay: After addressing all comments, I will manually create packaging-test images to test the developments made here on https://github.com/citusdata/tools/pull/167

scripts/fetch_and_build_deb
```diff
@@ -49,15 +49,25 @@ fi
 # populate variables from packaging metadata file
 # shellcheck source=/dev/null
 source /buildfiles/pkgvars
+# populate pg versions to be used for packaging.
+# supported-potgres file is derived file from postgres-matrix.yml file
+source /buildfiles/supported-postgres 2>/dev/null
```
hanefi: This will fail if the file does not exist. Redirecting stderr does not prevent nonzero vales returned by `source`

gurkanindibay: image

hanefi: You can check exit code of the last command by `echo $?`

Alternatively, you can use `set -euo pipefail` in bash to have a similar setting on your terminal.

Not having any output on your screen does not guarantee that the command ran successfully.

gurkanindibay: You're right I added file existence check now
https://github.com/citusdata/packaging/blob/cc3368bd5400d7c3e7bd0ec59c56b46d1d7741d4/scripts/fetch_and_build_deb#L55

scripts/fetch_and_build_deb
```diff
@@ -49,15 +49,26 @@ fi
...
 versioning="${versioning:-simple}"
+# Fetch pg release and nightly versions from supported-postgres file which is originated from postgres-matrix.yml file
+# If this file is not found, releasepg and nightlypg parameters from pkgvars are used for defining pg versions
+pg_release_versions="${release_versions:-${releasepg}}"
+pg_nightly_versions="${nightly_versions:-${nightlypg}}"
+
+
+echo "Postgres versions:"
+echo "Release Versions :${pg_release_versions}"
+echo "Nightly Versions :${pg_nightly_versions}"
```
hanefi:
```suggestion
echo "Release Versions: ${pg_release_versions}"
echo "Nightly Versions: ${pg_nightly_versions}"
```

scripts/fetch_and_build_deb
```diff
@@ -49,15 +49,26 @@ fi
...
 pkgname="${deb_pkgname:-${pkgname}}"
 hubproj="${hubproj:-${pkgname}}"
 nightlyref="${nightlyref:-master}"
-releasepg="${releasepg:-11,12,13}"
-nightlypg="${nightlypg:-${releasepg}}"
 versioning="${versioning:-simple}"
+# Fetch pg release and nightly versions from supported-postgres file which is originated from postgres-matrix.yml file
+# If this file is not found, releasepg and nightlypg parameters from pkgvars are used for defining pg versions
+pg_release_versions="${release_versions:-${releasepg}}"
+pg_nightly_versions="${nightly_versions:-${nightlypg}}"
```
hanefi, line 66:
```suggestion
pg_release_versions="${release_versions:-$releasepg}"
pg_nightly_versions="${nightly_versions:-$nightlypg}"
```

scripts/fetch_and_build_deb
```diff
@@ -49,15 +49,26 @@ fi
 # populate variables from packaging metadata file
 # shellcheck source=/dev/null
 source /buildfiles/pkgvars
+# populate pg versions to be used for packaging.
+# supported-potgres file is derived file from postgres-matrix.yml file
```
hanefi:
```suggestion
# Read PostgreSQL versions from file.
# supported-postgres file is derived file from postgres-matrix.yml file by citus_package.
```

hanefi (approved):
Minor formatting suggestions that should be applied to all the files that are generated from the templates.

scripts/fetch_and_build_deb
gurkanindibay: I think we should keep this since it is better using bash parameters with curly braces

PR #739, opened 2021-08-25
All citus unit tests with postgres matrix

gurkanindibay:
This is the test branch which includes postgres matrix. releasepg and nightlypg is removed from branches

### DIFF
--- a/postgres-matrix.yml
+++ b/postgres-matrix.yml
@@ -1,0 +1,16 @@
+name: Postgres Version Matrix
+project_name: citus # alternatives: citus, citus-enterprise, pg-auto-failover, pg-auto-failover-enterprise
+# There will be one configuration like this for each project in packaging repo
+# i.e. in all-citus, all-enterprise,all-pgautofailover,all-pgautofailover-enterprise, pgxn-citus etc.
+version_matrix:
+  - 8.0:
+      postgres_versions: [10, 11]
+  - 9.0:
+      postgres_versions: [11, 12]
+  - 9.5:
+      postgres_versions: [11, 12, 13]
+  # If 10.0 is released, since it is between 10.1 and 9.5, 9.5 support will be effective for 10.0.x releases
+  - 10.1:
+      postgres_versions: [12, 13]
+  - 10.2:
+      postgres_versions: [12, 13, 14]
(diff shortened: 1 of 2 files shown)


postgres-matrix.yml
```diff
@@ -0,0 +1,14 @@
+name: Postgres Version Matrix
+project_name: citus # alternatives: citus, citus-enterprise, pg-auto-failover, pg-auto-failover-enterprise
+# There will be one configuration like this for each project in packaging repo
+# i.e. in all-citus, all-enterprise,all-pgautofailover,all-pgautofailover-enterprise, pgxn-citus etc.
```
hanefi: We usually have comments at the end of a line, or right before it starts.

I think that the comments here are not really needed as this branch is concerned only with citus, and not other possible values of `project_name`
```suggestion
project_name: citus
```

gurkanindibay: Fixed

postgres-matrix.yml
```diff
@@ -0,0 +1,13 @@
...
+project_name: citus
+
+version_matrix:
+  - 8.0:
+      postgres_versions: [10, 11]
+  - 9.0:
+      postgres_versions: [11, 12]
+  - 9.5:
+      postgres_versions: [11, 12, 13]
+  # If 10.0 is released, since it is between 10.1 and 9.5, 9.5 support will be effective for 10.0.x releases
```
hanefi, line 11: I think it makes more sense to have an entry for 10.0, instead of having a comment that is actually longer than the alternative.

```suggestion
  - 10.0:
      postgres_versions: [11, 12, 13]
```

postgres-matrix.yml
```diff
@@ -0,0 +1,13 @@
...
+  - 10.1:
+      postgres_versions: [12, 13]
```
hanefi, line 13: We already created 10.2 packages. Can we add that to the matrix as well?

```suggestion
      postgres_versions: [12, 13]
  - 10.2:
      postgres_versions: [12, 13, 14]
```

hanefi (changes requested):

postgres-matrix.yml
gurkanindibay: I added it just to be easily understood no need to add an entry

hanefi: ok. your call

hanefi (approved):

PR #742, opened 2021-08-27
Citus nightly pipeline development

gurkanindibay:
- [ ] There was a small bug which was preventing package push into citusdata/tools#173. After its completion tag name should be changed with the new tag

### DIFF
--- a/.github/workflows/build-citus-nightlies.yml
+++ b/.github/workflows/build-citus-nightlies.yml
@@ -1,10 +1,12 @@
-name: Buildand publish citus nightly package
+name: Build and publish citus nightly package
 
 env:
-  MAIN_BRANCH: "all-citus"
+  MAIN_BRANCH: "develop"
   PACKAGE_CLOUD_REPO_NAME: "citusdata/community-nightlies"
-  PACKAGE_CLOUD_API_TOKEN: ${{ secrets.PACKAGE_CLOUD_API_TOKEN }}
-  GH_TOKEN: ${{ secrets.GH_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build-citus-nightlies.yml
```diff
@@ -0,0 +1,71 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Clone tools branch
+        run: git clone -b pc_invalid_repo_name --depth=1  https://github.com/citusdata/tools.git tools
+
+      - name: Clone all-citus branch
+        run: git clone -b all-citus --depth=1  https://github.com/citusdata/packaging.git packaging
```
onurctirtir: what about enterprise ?

gurkanindibay: I can add matrix structure and add enterprise in it but I prefer another PR to make nightly quickly work

gurkanindibay: @onurctirtir having another pipeline is more accurate way to handle enterprise . I will open another PR for it.
But I need this pipeline to test 10.2.0 release on Github Actions just like we release 10.2.0

hanefi: Please open an issue to make sure we do not forget about enterprise nightly images

hanefi (approved):
PR looks like it's ok when I read it. Sadly no easy way to run it.

Please open an issue for a similar cron job for enterprise packages as well.

I think the best way to see if this cron works properly is merging it and checking the outputs in the following days.

PR #745, opened 2021-09-03
Moves packaging image creation into Github Actions

gurkanindibay:


ci/push_images
```diff
@@ -12,12 +12,12 @@ badusage=64
...
 nprocs="${1:-1}"
 
-if [[ $TRAVIS_PULL_REQUEST != "false" ]] || [[ $TRAVIS_BRANCH != "develop" ]] ; then
+if [[ ${CURRENT_BRANCH} != "develop" ]] ; then
     echo "PR is not merged to develop, not pushing to DockerHub"
     exit 0;
 fi
 
-echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+echo "${DOCKERHUB_PASSWORD}" | docker login -u "${DOCKERHUB_USERNAME}" --password-stdin
```
gurkanindibay: We store dockerhub parameters in Github actions with the names above

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Test packages using built images
+        run: |
+          python -m  tools.packaging_automation.citus_package \
+          --gh_token "${GH_TOKEN}" \
+          --packaging_docker_platform "${{ matrix.TARGET_PLATFORM }}" \
+          --build_type 'release' \
+          --secret_key "${PACKAGING_SECRET_KEY}" \
+          --passphrase "${PACKAGING_PASSPHRASE}" \
+          --output_dir "$(pwd)/packages/" \
+          --input_files_dir "$(pwd)"
```
gurkanindibay: This section is performing the test. We are executing out tests on all-citus branch, i.e. performing tests on citus app.

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Build images
+        run: |
+          git checkout -- dockerfiles
+          ./update_images
+
+      - name: Clone all-citus branch for test
+        run: git clone -b all-citus --depth=1  https://github.com/citusdata/packaging.git packaging
+
+      - name: Clone tools repo for test
+        run: git clone -b develop_tests --depth=1  https://github.com/citusdata/tools.git packaging/tools
```
gurkanindibay: The branch name will be changed to related tag name after closing the PR below
https://github.com/citusdata/tools/pull/177

hanefi:
Needs a rebase

ci/push_images
hanefi: Are they not secrets? How does this work?

gurkanindibay: These are environment variables. I missed adding as environment variables added them

.github/workflows/build-citus-nightlies.yml
```diff
@@ -0,0 +1,71 @@
...
+    name: Build package
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - el/7
+          - el/8
+          - ol/7
+          - debian/stretch
```
hanefi: What about debian bullseye?

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,73 @@
+name: Build Package
+
+env:
+  GH_TOKEN: "${{ secrets.GH_TOKEN }}"
+  PACKAGING_SECRET_KEY: "${{ secrets.PACKAGING_SECRET_KEY }}"
+  PACKAGE_CLOUD_REPO_NAME: "sample"
```
hanefi: Do we want to change this?

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Install package dependencies
+        run: sudo apt-get update && sudo apt-get install libcurl4-openssl-dev libssl-dev python3-testresources
+
+      - name: Install wheel for el/8
+        if: matrix.TARGET_PLATFORM  == 'centos,8'
+        run: python -m pip install wheel
+
+      - name: Update and check dockerfiles
+        run: |
+          ./update_dockerfiles
```
hanefi: I have 2 concerns here.
1. We do not need to run this script for every OS in the matrix. One is enough.
2. Since we do not fail fast, the test outputs are not clear if this step fails.

Can we somehow change the structure here a little bit?

ci/push_images
```diff
@@ -12,12 +12,12 @@ badusage=64
...
 nprocs="${1:-1}"
 
-if [[ $TRAVIS_PULL_REQUEST != "false" ]] || [[ $TRAVIS_BRANCH != "develop" ]] ; then
+if [[ ${CURRENT_BRANCH} != "develop" ]] ; then
     echo "PR is not merged to develop, not pushing to DockerHub"
     exit 0;
 fi
 
-echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+echo "${DOCKERHUB_PASSWORD}" | docker login -u "${DOCKERHUB_USER_NAME}" --password-stdin
```
hanefi, line 20: You could move this to CI configs. Even better you can use docker actions to login

.github/workflows/build-package.yml
gurkanindibay: copy paste error. Removed

.github/workflows/build-package.yml
gurkanindibay: 1. Created a job to assure this step executes once
2. https://github.com/citusdata/packaging/blob/56870d2c26e47b9b76a73701a953761e1f9c39f4/update_dockerfiles#L4
Script is failing fast and Github Actions fails fast by default so

ci/push_images
gurkanindibay: I think we should keep this to be able to use this scripts manually. It does not seem right to me to break the consistency in the script. Docker login is the part of the logic that we packed into script. Either we should remove all the commands and put them in GH Actions or keep the script as is

.github/workflows/build-citus-nightlies.yml
gurkanindibay: Removed pipeline since it is the duplicate of build-citus-community-nightlies.yml

ci/push_images
hanefi: In my local machine, I would not want to store my username and password in env vars that can be compromised.

Docker login should be decoupled with these scripts, or they should be optional.

However, I am ok with this as long as you open a new issue and track this request.

update_images
```diff
@@ -14,33 +14,33 @@ nprocs="${1:-1}"
...
-    IFS=',' read -r os release <<< "$line"
 
-    if [[ "${os}" = 'debian' ]] || [[ "${os}" = 'ubuntu' ]]; then
-        tag="${os}-${release}-all"
+IFS=',' read -r os release <<< "${TARGET_PLATFORM}"
+
+if [[ "${os}" = 'debian' ]] || [[ "${os}" = 'ubuntu' ]]; then
+    tag="${os}-${release}-all"
+    args+="build -t citus/packaging:${tag} -f ${dockerfiles_dir}/${tag}/Dockerfile .\n"
+elif [[ "${os}" = 'centos' ]] || [[ "${os}" = 'fedora' ]] || [[ "${os}" = 'oraclelinux' ]]; then
```
hanefi: We do not create fedora packages, and I think it is safe to remove them from here

.github/workflows/build-package.yml
hanefi: What about Line 34 (`fail-fast: false`) then? Does that not mean anything?

https://github.com/citusdata/packaging/pull/745/files#diff-c37cf2bffb2cde25cbcedc64387a46a1b752a13b0f310ecc68058b87889a3ac0R34

ci/push_images
gurkanindibay: https://github.com/citusdata/packaging/issues/828

.github/workflows/build-package.yml
gurkanindibay: https://www.edwardthomson.com/blog/github_actions_6_fail_fast_matrix_workflows.html
It is about matrix which enables all the other jobs continue in the matrix even if one of them fails

update_images
gurkanindibay: Fixed

update_images
```diff
@@ -14,33 +14,33 @@ nprocs="${1:-1}"
 
 declare args
 
-while read -r line; do
-    IFS=',' read -r os release <<< "$line"
 
-    if [[ "${os}" = 'debian' ]] || [[ "${os}" = 'ubuntu' ]]; then
-        tag="${os}-${release}-all"
+IFS=',' read -r os release <<< "${TARGET_PLATFORM}"
```
hanefi: why do we no longer have a loop here? I do not like the fact that a script called `update_images` updates the images for a single platform.

It was easier to use this script when we could update all images that are listed in `os-list.csv`

gurkanindibay: It is very error prone. I had very long build times when I first try to move this script. Then I realized that I didn't set  the parameters properly and built all the images for just one image type
Therefore I can update the script name into update_image if you want since this script will never be used for all the images in the pipeline

update_image
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+#This script create/update image by getting the the parameter of os release pair splitted with comma
```
hanefi:
```suggestion
# This script builds a docker image that is used for packaging.
#
# This script requires $TARGET_PLATFORM env to contain a supported platform name.
```

update_image
```diff
@@ -0,0 +1,50 @@
...
+
+badusage=64
+
+
+nprocs="${1:-1}"
+
+declare args
+
+
+IFS=',' read -r os release <<< "${TARGET_PLATFORM}"
```
hanefi, line 24: Can we at least loop through all supported platform names when ${TARGET_PLATFORM} is empty?

This way, we will be able to update all images locally and have a more robust development environment.

hanefi (approved):
Please consider my comments before merging

update_image
gurkanindibay: That is why I lost lots of time. I didn't notice that getting the full list. It tried to execute all the images.
I can write another script for that. 
https://github.com/citusdata/packaging/issues/830

PR #749, opened 2021-09-09
Bump to pg_auto_failover 1.6.2

gurkanindibay:


onurctirtir (approved):
I assume Dimitri is okay with dropping xenial support.

PR #763, opened 2021-09-14
Adds github actions pipeline for pg-auto-failover

gurkanindibay:


hanefi:
I compared `.tavis.yml` contents with the new `build-package.yml` and I saw some missing pieces.

1. Why did we drop nightly package creation? I saw that we have a similar issue with Citus packages as well. How do we plan to create nightly packages after this PR is merged?
2. Do we not require docker logins to increase docker pull limits here? We used to have a step in `before_install` section: `echo "${DOCKER_PASS}" | docker login --username ${DOCKER_USER} --password-stdin`
3.

gurkanindibay:
> I compared `.tavis.yml` contents with the new `build-package.yml` and I saw some missing pieces.
> 
> 1. Why did we drop nightly package creation? I saw that we have a similar issue with Citus packages as well. How do we plan to create nightly packages after this PR is merged?
> 2. Do we not require docker logins to increase docker pull limits here? We used to have a step in `before_install` section: `echo "${DOCKER_PASS}" | docker login --username ${DOCKER_USER} --password-stdin`

My next task is creating nightlies. I will create another pipeline to handle four flows:
citus,citus-enterprise, pgaf and pgaf enterprise
Adding will increase our bandwith good catch I added login step

.github/workflows/build-package.yml
```diff
@@ -31,31 +35,36 @@ jobs:
...
       - name: Install package dependencies
-        run: sudo apt install libcurl4-openssl-dev libssl-dev
+        run: sudo apt-get update && sudo apt-get install libcurl4-openssl-dev libssl-dev python3-testresources
 
       - name: Install python requirements
         run: python -m pip install -r tools/packaging_automation/requirements.txt
 
+      # This step is to fetch the images unanonymously to have higher bandwidth
+      - name: DockerHub Login
+        run: docker login -u "${DOCKERHUB_USER_NAME}" -p "${DOCKERHUB_PASSWORD}"
```
hanefi: I see that the tests pass, but I do not understand how. Where do we define these env variables? They are not in env section of this yaml, and they do not seem to be secrets. Can you show me where they are defined? Does it make sense to make them secrets?

gurkanindibay: yes it makes sense. we may define username openly but password should be secret. But it seems reasonable to use username as secret for two reasons 
1. It seems it is not secret but we make attackers life easier by defining it here.
2. I define only one parameter and use it everywhere which I think is a good practice

https://github.com/citusdata/packaging/settings/secrets/actions
You can access these parameters from above address

gurkanindibay: I checked it I forgot to define secrets as env variables it is interesting not having error by the way. :(

.github/workflows/build-package.yml
```diff
@@ -31,31 +35,39 @@ jobs:
...
+        run: sudo apt-get update && sudo apt-get install libcurl4-openssl-dev libssl-dev python3-testresources
 
       - name: Install python requirements
         run: python -m pip install -r tools/packaging_automation/requirements.txt
 
+      # This step is to fetch the images unanonymously to have higher bandwidth
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USER_NAME }}
```
hanefi, line 50: I do not think that it makes sense to keep a user name as a secret. The docker hub account is already publicly available. I do not think that this is really necessary

hanefi: However it is really nice to be able to store the docker hub account name in a single place. I am ok with the change here now.

hanefi (approved):

PR #767, opened 2021-09-14
Adds github actions build pipeline for redhat topn

gurkanindibay:


hanefi:
I think we should merge `debian-topn` and `redhat-topn` branches and decrease the amount of effort needed in here.

Similarly for hll, we can create `all-hll`.

I can work on merging the branches, as I have attempted it already and was quite successful.

hanefi (approved):

PR #772, opened 2021-09-16
Bump pgxn to version 10.2.0

gurkanindibay:


hanefi:
We still have a prerequisite for `12.0.0`. Should we change that?

It would be nice if we can somehow require `>= 12.0.0`

https://github.com/citusdata/packaging/blob/a70b5ae7c972cfeeb488a67ea375873691dd5e79/META.json#L16-L21

hanefi:
> We still have a prerequisite for `12.0.0`. Should we change that?

Apparently not. It is ok.

See the spec here: https://pgxn.org/meta/spec.txt

hanefi (approved):

PR #773, opened 2021-09-16
Remove encryption from Citus Enterprise PackageCloud packages

### LINKED ISSUE #774: Add unencrypted support for citus_package
Now packaging_secret_key and packaging_passphrase is mandatory for citus package and for packagecloud it should be unencrypted. if it is unencrypted then these parameters should not be taken. It increases attack surface. Remove the encryptioon parameter need when packages to be released is unencrypted

gurkanindibay:
I tried removing   PACKAGING_SECRET_KEY and  PACKAGE_ENCRYPTION_KEY variables but now script needs both of them. I opened issue #774 to remove these parameters


hanefi:
@gurkanindibay please bump the fancy version to 2, and update versions to create a new package name

hanefi (approved):

PR #776, opened 2021-09-17
Removes encryption from pgaf-enterprise

gurkanindibay:
I have performed the changes we applied on citus-enterprise into pgaf enterprise I also need to recreate the packages after approval


hanefi (approved):
This change is ok assuming you will create another PR for fancy version bump.

PR #778, opened 2021-09-17
Bump to pg_auto_failover_enterprise 1.6.2-2

gurkanindibay:


pkgvars
```diff
@@ -2,6 +2,6 @@ rpm_pkgname=pg-auto-failover-enterprise
 deb_pkgname=auto-failover-enterprise
 hubproj=citus-ha
 pkgdesc='Postgres extension for automated failover and high-availability'
-pkglatest=1.6.2-1
+pkglatest=1.6.2-2
 releasepg=10,11,12,13
```
hanefi: Can we also add 14 here ?

```suggestion
releasepg=10,11,12,13,14
```

pg-auto-failover-enterprise.spec
```diff
@@ -371,6 +371,9 @@ rm -f %{pginstdir}/bin/pg_autoctl
 
 
 %changelog
+* Fri Sep 17 2021 - Gurkan Indibay <gindibay@microsoft.com> 1.6.2-2
+- Official 1.6.2 release of Pg Auto Failover Enterprise
```
hanefi: We can have a better description here for future reference.

```suggestion
- Removes encryption from packages
```

gurkanindibay: Fixed

pkgvars
gurkanindibay: Fixed

hanefi (approved):

PR #780, opened 2021-09-17
Bump to citus_enterprise 10.1.3

gurkanindibay:


pkgvars
```diff
@@ -1,7 +1,7 @@
 pkgname=citus-enterprise
 pkgdesc='Citus Enterprise'
-pkglatest=10.2.0.citus-2
-releasepg=12,13,14
-nightlypg=12,13,14
+pkglatest=10.1.3.citus-1
+releasepg=12,13
+nightlypg=12,13
```
hanefi: We can consider keeping the nightlypg as is.

```suggestion
nightlypg=12,13,14
```

hanefi (approved):

PR #791, opened 2021-09-24
Bump to citus_enterprise 10.2.1

gurkanindibay:


hanefi (approved):
I do not think that it was necessary to update the changelog length by hand in 3220d22. However I am ok with it as it seems harmless.

PR #793, opened 2021-09-27
Adds debian bullseye packaging image among other supported operating system images

gurkanindibay:
Installed with using packagecloud installation scripts 
Need to change deb.sh for community nightlies and enterprise
image

### DIFF
--- a/os-list.csv
+++ b/os-list.csv
@@ -2,6 +2,7 @@ centos,8
 centos,7
 debian,buster
 debian,stretch
+debian,bullseye
 oraclelinux,8
 oraclelinux,7
 oraclelinux,6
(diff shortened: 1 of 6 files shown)


os-list.csv
```diff
@@ -2,6 +2,7 @@ centos,8
 centos,7
 debian,buster
 debian,stretch
+debian,bullseye
```
gurkanindibay, line 5: After adding this line I only executed ./update_dockerfiles

onurctirtir: great

onurctirtir (approved):
> Need to change deb.sh for community nightlies and enterprise

True. Plus, I'm wondering should we make some changes in `citus_package` to start building packages for this distro, maybe a small matrix change or something like that .

gurkanindibay:
> > Need to change deb.sh for community nightlies and enterprise
> 
> True. Plus, I'm wondering should we make some changes in `citus_package` to start building packages for this distro, maybe a small matrix change or something like that .

Yes I will send other PR's to add this support for pipelines

PR #795, opened 2021-09-27
Adds debian bullseye support into nightly pipeline

gurkanindibay:


onurctirtir (approved):
we are okay with travis to fail

PR #796, opened 2021-09-27
Remove installation scripts for PG 9.6

hanefi:
We stopped creating and maintaining packages for PG9.6 for some time
now. It makes sense to remove those files and decrease the technical
debt we have in the codebase.


gurkanindibay:
Below is download details for postgres 9.6 Still being downloaded so I think we should not remove these scripts
[package_cloud_download_details.csv](https://github.com/citusdata/packaging/files/7234720/package_cloud_download_details.csv)

onurctirtir:
> Below is download details for postgres 9.6 Still being downloaded so I think we should not remove these scripts
> [package_cloud_download_details.csv](https://github.com/citusdata/packaging/files/7234720/package_cloud_download_details.csv)

I came here to just approve the pr but didn't know this, thank you Gurkan.
So then it makes sense to not remove those download scripts @hanefi ?

hanefi:
I do not think that we actually have any data on whether these scripts are used by anyone. However I guess we can close this PR for now.

PR #801, opened 2021-09-27
Adds debian bullseye support into citus-enterprise

gurkanindibay:


debian/compat
```diff
@@ -1 +1 @@
-9
```
onurctirtir, line 1: why this change ?

onurctirtir (approved):

debian/compat
gurkanindibay: https://github.com/citusdata/packaging/runs/3720189231?check_suite_focus=true#step:6:403
Gave warning in packaging process
So changed in all of them

PR #806, opened 2021-09-28
Adds ignore for useless-provides for rpm in pgaf

gurkanindibay:
When building nightlies we saw the error below
https://github.com/citusdata/packaging/runs/3738628202?check_suite_focus=true
When I look into the error, I realized that this same error has been addes ignore list for citus
packaging/rpmlintrc

Line 2 in 383c0a7

 addFilter("E: useless-provides .*"); 

When I look for this error since the same package is shown multiple times it is not an error and it is ignorable as defined below
rpm-software-management/rpmlint#427


onurctirtir (approved):
thank you for investigation and fixing this!

PR #807, opened 2021-09-28
Adds ignore for useless-provides for rpm in pgaf enterprise

gurkanindibay:
When building nightlies we saw the error below
https://github.com/citusdata/packaging/runs/3662202754?check_suite_focus=true
When I look into the error, I realized that this same error has been addes ignore list for citus
 https://github.com/citusdata/packaging/blob/383c0a7a575cc198787ab35688a8c85a904b5305/rpmlintrc#L2
When I look for this error since the same package is shown multiple times it is not an error and it is ignorable as defined below
https://github.com/rpm-software-management/rpmlint/issues/427


onurctirtir (approved):
thank you for investigation and fixing this!

PR #812, opened 2021-09-30
Uses pg14 stable changed tools

gurkanindibay:


hanefi:
I think you could have done this change when creating the next package release. Do you plan to update all affected branches in this repo one by one?

gurkanindibay:
> I think you could have done this change when creating the next package release. Do you plan to update all affected branches in this repo one by one?

Yes after tagging I will change all of them. Yes you're right I can change that time but making now seems more reasonable to me since changes at release time should be related with only release issues I think

gurkanindibay:
No need so removed it since handled it with https://github.com/citusdata/packaging/pull/820

PR #813, opened 2021-10-01
Remove travis config from develop

hanefi:
We stopped using Travis and the transition to Github Actions is not complete.

I will remove Travis from required checks, once I get an approval in this PR.


gurkanindibay (changes requested):
Already .travis.yml  is removed in this PR with the GH Actions replacement pipelines
No need for this PR
https://github.com/citusdata/packaging/pull/745

PR #814, opened 2021-10-01
Updates pg version from 13 to 14

gurkanindibay:
[test-centos7.txt](https://github.com/citusdata/packaging/files/7267627/test-centos7.txt
[test-focal.txt](https://github.com/citusdata/packaging/files/7267676/test-focal.txt)
)


hanefi (approved):
Looks good. While testing, I found that our intermittent problems with RHEL7 is no more apparently.

PR #848, opened 2021-11-01
Adds ol/8 support for packages for pg-cron

gurkanindibay:


postgres-matrix.yml
```diff
@@ -0,0 +1,6 @@
+name: Postgres Version Matrix
+project_name: pg-cron
+
+version_matrix:
+  - 1.4.1:
+      postgres_versions: [10, 11, 12, 13, 14]
```
onurctirtir, line 6: I wonder how we were building pg_cron packages without this file before ?

onurctirtir (approved):

postgres-matrix.yml
gurkanindibay: if this file is missing and releasepg is non-empty, we were using releasepg value

PR #859, opened 2021-11-13
Adds notice file into debian packages

gurkanindibay:


debian/docs
```diff
@@ -0,0 +1 @@
+debian/NOTICE
```
hanefi, line 1: If you decide to introduce this file, you need to specify more items. Not having this file, makes `dh_installdocs` fall back to the default behaviour.

> By default, it will include all existing files in the top-level source directory that are called BUGS, README*, TODO etc.

You should at least add README.md files, and make sure we do not stop adding any documentation when we are trying to add the notice file

debian/NOTICE
```diff
@@ -0,0 +1,99 @@
+NOTICES AND INFORMATION
```
hanefi, line 1: If we need to add the NOTICE file to the citus repo, can we use that file and _not_ create another copy in here?

debian/docs
gurkanindibay: But we don't have a README file for now. And it is out of the scope of the task. If we need a README file in the future, we can open a task, proritize it and after that I can happily add this file and add into the docs file.

debian/NOTICE
gurkanindibay: Actually I don't like the idea. I understand you are trying to enforce DRY principle but it is not a good practice to get an online file during packaging process which may cause flakiness.

.github/workflows/build-package.yml
```diff
@@ -81,6 +81,6 @@ jobs:
           version=${pkglatest%.citus-1}
           echo ::set-output name=version::${version}
       - name: Trigger package tests
-        run: gh workflow run "citus-package-tests" --repo citusdata/tools -f prj_ver="${{ steps.set_version.outputs.version }}"
+        run: gh workflow run "citus-package-tests" --repo citusdata/tools -f project_version="${{ steps.set_version.outputs.version }}"
```
hanefi: Can you split this into several lines?

debian/NOTICE
hanefi: Each version of Citus can potentially have different NOTICE files. How do you plan to maintain different versions of these files?

I do not think that we can move forward without having a working plan for this problem.

hanefi: I went through all Citus versions >= 9.2 and the NOTICE file is correct for all of those.

I am ok with keeping the NOTICE file here, for now.

.github/workflows/build-package.yml
hanefi:
```suggestion
        run: gh workflow run "citus-package-tests" \
             --repo citusdata/tools \
             -f prj_ver="${{ steps.set_version.outputs.version }}"```

gurkanindibay:
@hanefi  #860 opened to keep NOTICE file up to date

hanefi (approved):

.github/workflows/build-package.yml
gurkanindibay: Fixed

PR #863, opened 2021-11-25
Fixes enterprise and nightlies repo installations

gurkanindibay:
It seems that packagecloud started storing gpg keys in armored style and when adding as gpg keydearmor operation is needed to perform to convert it to binary and then store it as keyring
Armor
https://www.techopedia.com/definition/23150/ascii-armor
dearmor
https://www.linuxjournal.com/article/8732

### DIFF
--- a/community-nightlies/deb.sh
+++ b/community-nightlies/deb.sh
@@ -273,6 +273,8 @@ main ()
 
   echo -n "Importing Citus Data gpg key... "
   # import the gpg key
+  # below command decodes the ASCII armored gpg file (instead of binary file)
+  # and adds the unarmored gpg key as keyring
   curl -L "${gpg_key_url}" 2> /dev/null | apt-key add - &>/dev/null
   echo "done."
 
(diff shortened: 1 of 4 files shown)


community/deb.sh
```diff
@@ -273,6 +273,8 @@ main ()
 
   echo -n "Importing Citus Data gpg key... "
```
hanefi: For the sake of consistency, can we do the following change?

```suggestion
  echo -n "Importing Citus Data Community gpg key... "
```

community-nightlies/deb.sh
```diff
@@ -273,6 +273,8 @@ main ()
 
   echo -n "Importing Citus Data gpg key... "
```
hanefi: For the sake of consistency, can we do the following change?

```suggestion
  echo -n "Importing Citus Data Community nightlies gpg key... "
```

hanefi (changes requested):

hanefi (approved):

PR #864, opened 2021-11-26
Fixes ol/7 package build problem

gurkanindibay:
Fixed the following problem . Only executed yum update before the failed command
https://github.com/citusdata/packaging/runs/4321343936?check_suite_focus=true

### DIFF
--- a/dockerfiles/centos-7-pg11/Dockerfile
+++ b/dockerfiles/centos-7-pg11/Dockerfile
@@ -40,6 +40,7 @@ RUN if [[ centos   == oraclelinux ]] && [[ 7 == 7 ]]; then yum install -y wget \
 # install build tools and PostgreSQL development files
 RUN ( yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
     && [[ -z "epel-release centos-release-scl-rh" ]] || yum install -y epel-release centos-release-scl-rh) \
+    && yum update -y \
     && yum groupinstall -y 'Development Tools' \
     && yum install -y \
         curl \
--- a/templates/Dockerfile-rpm.tmpl
+++ b/templates/Dockerfile-rpm.tmpl
@@ -40,7 +40,6 @@ RUN if [[ %%os%%   == oraclelinux ]] && [[ %%release%% == 7 ]]; then yum install
 # install build tools and PostgreSQL development files
 RUN ( yum install -y https://%%rpm_url%% \
     && [[ -z "%%extra-repositories%%" ]] || yum install -y %%extra-repositories%%) \
-    && yum update -y \
     && yum groupinstall -y 'Development Tools' \
     && yum install -y \
         curl \
--- a/templates/Dockerfile-rpm.tmpl
+++ b/templates/Dockerfile-rpm.tmpl
@@ -40,6 +40,7 @@ RUN if [[ %%os%%   == oraclelinux ]] && [[ %%release%% == 7 ]]; then yum install
 # install build tools and PostgreSQL development files
 RUN ( yum install -y https://%%rpm_url%% \
     && [[ -z "%%extra-repositories%%" ]] || yum install -y %%extra-repositories%%) \
+    && yum update -y \
     && yum groupinstall -y 'Development Tools' \
     && yum install -y \
         curl \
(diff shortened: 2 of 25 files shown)


dockerfiles/centos-7-pg11/Dockerfile
```diff
@@ -38,7 +38,8 @@ RUN if [[ centos   == oraclelinux ]] && [[ 7 == 7 ]]; then yum install -y wget \
 
 
 # install build tools and PostgreSQL development files
-RUN ( yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
+RUN yum update -y \
+    && ( yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
```
onurctirtir: Normally, we would run that command as in below suggestion since the pr title claims that it only addresses to ol/7, consider my suggestion for the template dockerfile:
```suggestion
RUN if [[ $osname == oraclelinux ]] && [[ $osver == 7 ]]; then yum update -y

RUN ( yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
```

However, I guess there is no harm doing that for all rpm's, right ?

gurkanindibay: ### I thinks there is a potential for all of the packages since new repo is being added. I changed the order of the yum update -y to cover the two repos  being added below

templates/Dockerfile-rpm.tmpl
```diff
@@ -40,6 +40,7 @@ RUN if [[ %%os%%   == oraclelinux ]] && [[ %%release%% == 7 ]]; then yum install
 # install build tools and PostgreSQL development files
 RUN ( yum install -y https://%%rpm_url%% \
     && [[ -z "%%extra-repositories%%" ]] || yum install -y %%extra-repositories%%) \
+    && yum update -y \
```
hanefi: We already have the same command in the first `RUN` on line 4:

```
RUN yum -y update
```

Why do we need to update all the packages one more time? Is this really necessary?

You may have issues if you use some caches for the intermediate steps. That can be solved by adding `--no-cache` option when using `docker build`

dockerfiles/centos-7-pg11/Dockerfile
hanefi: I have several remarks:
1. `yum` does not really need explicit commands to update the cache.
2. If you wish to install the latest information on all available packages on a repo, you should run `yum check-update` instead, and make sure you consume the return value properly
3. `yum update` is similar to running `apt update; apt upgrade` as it installs all updates to currently installed packages.
4. it does not really make sense to me to update packages for ol7 and not do that on other distros.

After you go over these comments. can you update the PR and leave a comment on why you want these changes? Please focus on why, and skip what or how as your code changes should speak for itself.

templates/Dockerfile-rpm.tmpl
gurkanindibay: We are adding new repositories so yum update is needed after adding new repo

hanefi: > yum update is needed after adding new repo

No it is not. We do not need to upgrade all the packages in the system after we install a new repo.

Do you see any `yum update` commands in our repo installation scripts?

Here is where we install the new repo: https://github.com/citusdata/packaging/blob/71e84ec5a2a038e770af45a1a53b65e1cd26e9f9/community/rpm.sh#L45

dockerfiles/centos-7-pg11/Dockerfile
onurctirtir: I agree with all, thanks Hanefi

gurkanindibay: I already focused on why in my first comment  '''I thinks there is a potential for all of the packages since new repo is being added.''' I think this is a good why. I 'm checking your comment about yum update by which you gave the answer for the why. Thanks for the input

templates/Dockerfile-rpm.tmpl
gurkanindibay: I see your point. You're right it seems that yum install <repo_name> updates caches but still there are some non-up to date packages.  I can not find any other alternatives other thant yum update. Yum check-update only check for updates if there is shows the results. So I understand that first I need to call yum check-update and if there are updates I will call yum update. But yum update performs these steps internally and if there is no update it just print that all packages are up-to-date. So solely calling yum update makes more sense for me

gurkanindibay:
Thanks to @hanefi since https://github.com/citusdata/packaging/pull/865 solved the problem, there is no need this PR

PR #865, opened 2021-11-26
Do not use caches on docker build

hanefi:
I added two new options to our docker build commands.

--pull : Always attempt to pull a newer version of the base update_image

We can have new patches, or updates to our base images. It is a nice
idea to keep them updated.

--no-cache : Do not use cache when building the image

We can have some outdated package versions, if we use cached
intermediate steps. It is a good idea to run all the steps to make sure
we have up-to-date packages.

Related: https://github.com/citusdata/packaging/pull/864#discussion_r757590751


gurkanindibay (approved):

onurctirtir (approved):
thank you for detailed explanation

PR #873, opened 2022-02-01
Bump to citus 10.1.4

gurkanindibay:


citus.spec
```diff
@@ -43,8 +43,6 @@ make %{?_smp_mflags}
 # Install documentation with a better name:
 %{__mkdir} -p %{buildroot}%{pginstdir}/doc/extension
 %{__cp} README.md %{buildroot}%{pginstdir}/doc/extension/README-%{sname}.md
-%{__cp} NOTICE %{buildroot}%{pginstdir}/doc/extension/NOTICE-%{sname}
```
hanefi: Why did we remove this NOTICE file here?

pkgvars
```diff
@@ -1,4 +1,6 @@
 pkgname=citus
 pkgdesc='Citus (Open-Source)'
-pkglatest=10.2.3.citus-1
+pkglatest=10.1.4.citus-1
+releasepg=11,12,13
+nightlypg=12,13
```
hanefi: I do not think these fields make sense.

1. Do we need `releasepg` here?
2. Do we need `nightlypg` here?
3. PG11 support was dropped in `10.1` and we should not have that in this field here.

citus.spec
```diff
@@ -44,7 +44,6 @@ make %{?_smp_mflags}
 %{__mkdir} -p %{buildroot}%{pginstdir}/doc/extension
 %{__cp} README.md %{buildroot}%{pginstdir}/doc/extension/README-%{sname}.md
 %{__cp} NOTICE %{buildroot}%{pginstdir}/doc/extension/NOTICE-%{sname}
-
```
hanefi: Let's keep the empty line that makes the file easier to read.

citus.spec
gurkanindibay: Thanks for the catch. We have an open issue about updating templates. I need to update it manually until implementing this issue. I missed the manual check here

pkgvars
gurkanindibay: I missed that part as well. I deleted this part as well. Thanks for the comment

citus.spec
gurkanindibay: Fixed as you suggested

hanefi (approved):
Please go over the test outputs before merging this PR

PR #876, opened 2022-02-01
Bump pgxn to version 10.2.4

gurkanindibay:


pkgvars
```diff
@@ -1,4 +1,4 @@
 pkgname=citus
 pkgdesc='Citus (Open-Source)'
-pkglatest=10.2.3
+pkglatest=10.2.4
 releasepg=12,13,14
```
hanefi: `10.2` supports PG 11, 12, and 13.

14 is not supported.

Can we remove these references in pkgvars as they are not used?

gurkanindibay: No we can not we don't use postgres matrix here

hanefi: My comment above is not correct.

10.2 supports 12, 13 and 14.

10.1 supports 11, 12, and 13.

My bad.

However, I still think that we do not need the values here.

gurkanindibay: I removed these parameters

hanefi (approved):

PR #882, opened 2022-03-01
Fixes patch for ubuntu and non-working centos-8 mirrors

gurkanindibay:
While working on beta support for packages, I found out that docker build for ubuntu focal failed at the line below
https://github.com/citusdata/packaging/blob/351dd53fc1fd33f00af07e1b3af1e836914754f8/dockerfiles/ubuntu-focal-all/Dockerfile#L89
Additionally, centos repository mirrors were not working and I changed it with vault.epel.cloud
Removes nightlypg param from pgxn image which causes error after removal of releasepg and nightlypg parameters from pkgvars of pgxn
https://github.com/citusdata/packaging/blob/pgxn-citus/pkgvars

### DIFF
--- a/scripts/fetch_and_build_pgxn
+++ b/scripts/fetch_and_build_pgxn
@@ -44,7 +44,6 @@ source /buildfiles/pkgvars
 # set default values for certain packaging variables
 declare pkglatest # to make shellcheck happy
 hubproj="${hubproj:-${pkgname}}"
-nightlypg="${nightlypg:-${releasepg}}"
 
 if [ -z "${pkglatest}" ]; then
     echo "$0: pkgvars file must specify a value for pkglatest" >&2
--- a/templates/Dockerfile-rpm.tmpl
+++ b/templates/Dockerfile-rpm.tmpl
@@ -1,6 +1,11 @@
 # vim:set ft=dockerfile:
 FROM %%os%%:%%release%%
 
+RUN [[ %%os%% != centos ]] || [[ %%release%% != 8 ]] || ( \
+    cd /etc/yum.repos.d/ && sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-* \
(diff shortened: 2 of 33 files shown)


templates/Dockerfile-rpm.tmpl
```diff
@@ -1,6 +1,11 @@
 # vim:set ft=dockerfile:
 FROM %%os%%:%%release%%
 
+RUN [[ %%os%% != centos ]] || [[ %%release%% != 8 ]] || ( \
+    cd /etc/yum.repos.d/ && sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-* \
+    && sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.epel.cloud|g' /etc/yum.repos.d/CentOS-* \
+    )
+
```
hanefi, line 8: The current title of the PR and the change here seems unrelated. Can you elaborate?

You can consider moving this to another PR or updating the PR title, and description

scripts/fetch_and_build_pgxn
```diff
@@ -44,7 +44,6 @@ source /buildfiles/pkgvars
 # set default values for certain packaging variables
 declare pkglatest # to make shellcheck happy
 hubproj="${hubproj:-${pkgname}}"
-nightlypg="${nightlypg:-${releasepg}}"
```
hanefi, line 47: The current title of the PR and the change here seems unrelated. Can you elaborate?

You can consider moving this to another PR or updating the PR title, and description

hanefi (changes requested):

scripts/fetch_and_build_pgxn
gurkanindibay: Thanks for your comments. I changed the title and description. Could you check it again?

hanefi: title remains unchanged
description still does not cover why `nightlypg` was removed from the files here.

gurkanindibay: Apperantly, I didn't save the header :(. I changed it again.
Sorry for the loss of header

hanefi (approved):

PR #893, opened 2022-03-10
Adds beta support for packaging images

gurkanindibay:
packages needs to have beta keyword in minor version part. 
Ex: 
Deb: postgresql-14-citus-11.0-**beta**_11.0.0-**beta**.citus-2_amd64.deb 
Rpm: citus110_14_**beta**-11.0.0-**beta**.citus-2.el7.x86_64.rpm  

Additionally, to be able to test properly, I added test package creation support on `update_image` and `push_images` scripts

Post edits: I have test problems especially for centos releases. Recently, I have changed centos 8 repository, since redhat had cut the support. But this repository is much slower than the actual one so I got error nearly each run so I splitted each postgres execution in its own process


.github/workflows/build-package-test.yml
```diff
@@ -0,0 +1,79 @@
...
+
+env:
+  GH_TOKEN: "${{ secrets.GH_TOKEN }}"
+  PACKAGING_SECRET_KEY: "${{ secrets.PACKAGING_SECRET_KEY }}"
+  PACKAGE_ENCRYPTION_KEY: "${{ secrets.PACKAGE_ENCRYPTION_KEY }}"
+  PACKAGING_PASSPHRASE: "${{ secrets.PACKAGING_PASSPHRASE }}"
+  DOCKERHUB_PASSWORD: "${{ secrets.DOCKERHUB_PASSWORD }}"
+  DOCKERHUB_USER_NAME: "${{ secrets.DOCKERHUB_USER_NAME }}"
+  CURRENT_BRANCH: ${GITHUB_REF##*/}
+  TEST: true
```
gurkanindibay, line 11: I added a test parameter to be able to create test images from packaging

.github/workflows/build-package.yml
```diff
@@ -8,6 +8,7 @@ env:
   DOCKERHUB_PASSWORD: "${{ secrets.DOCKERHUB_PASSWORD }}"
   DOCKERHUB_USER_NAME: "${{ secrets.DOCKERHUB_USER_NAME }}"
   CURRENT_BRANCH: ${GITHUB_REF##*/}
+  TEST: false
```
gurkanindibay, line 11: Sets test parameter false for actual package creation to be able to create actual package

.github/workflows/build-package.yml
```diff
@@ -75,3 +76,5 @@ jobs:
 
       - name: Push images
         run: ./ci/push_images
+        env:
+          TARGET_PLATFORM: ${{ matrix.TARGET_PLATFORM }}
```
gurkanindibay, line 88: Previously push images was working for all images. Added target_platform parameter to be able to push just related image

ci/push_images
```diff
@@ -12,21 +12,29 @@ badusage=64
 
 nprocs="${1:-1}"
 
-if [[ ${CURRENT_BRANCH} != "develop" ]] ; then
+if [[ ${TEST} == "false" && ${CURRENT_BRANCH} != "develop" ]] ; then
```
gurkanindibay, line 15: packaging-test images are being pushed even if current-branch is not develop

ci/push_images
```diff
@@ -12,21 +12,29 @@ badusage=64
...
     echo "PR is not merged to develop, not pushing to DockerHub"
     exit 0;
 fi
 
+if [ "${TEST}" == "true" ]; then
```
gurkanindibay, line 20: Now this script supports both packaging-test and packaging images push

ci/push_images
```diff
@@ -12,21 +12,29 @@ badusage=64
...
+fi
+
 echo "${DOCKERHUB_PASSWORD}" | docker login -u "${DOCKERHUB_USER_NAME}" --password-stdin
 
 declare args
 
 while read -r line; do
     IFS=',' read -r os release <<< "$line"
-
+    if [ ! -z ${TARGET_PLATFORM} ] && [ ${line} != ${TARGET_PLATFORM} ]; then
```
gurkanindibay, line 32: When TARGET_PLATFORM is set, only image for this platform is pushed

dockerfiles/centos-7-pg10/scripts/fetch_and_build_rpm
```diff
@@ -58,6 +58,11 @@ nightlyref="${nightlyref:-master}"
 releasepg="${releasepg:-11,12,13}"
 nightlypg="${nightlypg:-${releasepg}}"
 versioning="${versioning:-simple}"
+if [[ "${pkglatest}" == *"beta"* ]]; then
```
gurkanindibay, line 65: Release type is set when beta is set in package version includes beta

dockerfiles/centos-7-pg10/scripts/fetch_and_build_rpm
```diff
@@ -166,8 +166,14 @@ case "${1}" in
     release)
         # add minor/major version to package name if using fancy versioning
         if [ "${versioning}" == 'fancy' ]; then
+
             infix=$(echo "${packageversion}" | grep -oE '^[0-9]+\.[0-9]+' | tr -d '.')
-            sed -i -E "1i %global pkginfix ${infix}" "${builddir}/${pkgname}.spec"
+            if [ "${release_type}" == "stable" ]; then
```
gurkanindibay, line 182: Beta keyword is added as package_prefix to be able to install with short name (102_beta)

update_image
```diff
@@ -20,12 +20,17 @@ nprocs="${1:-1}"
 
 declare args
 
+if [ "${TEST}" == "true" ]; then
```
gurkanindibay, line 23: Images with packaging-test name is being created with TEST environment variable

.github/workflows/build-package-test.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Build image for the target platform
+        run: |
+          git checkout -- dockerfiles
+          ./update_image
+        env:
+          TARGET_PLATFORM: ${{ matrix.TARGET_PLATFORM }}
+
+      - name: Clone tools repo for test
+        run: git clone -b v0.8.12 --depth=1  https://github.com/citusdata/tools.git tools
```
gurkanindibay: v0.8.12 tag is now on tools/beta_test branch. It will be moved after merge

.github/workflows/build-package.yml
```diff
@@ -64,7 +65,7 @@ jobs:
           TARGET_PLATFORM: ${{ matrix.TARGET_PLATFORM }}
 
       - name: Clone tools repo for test
-        run: git clone -b v0.8.8 --depth=1  https://github.com/citusdata/tools.git tools
+        run: git clone -b v0.8.12 --depth=1  https://github.com/citusdata/tools.git tools
```
gurkanindibay: v0.8.12 tag is now on tools/beta_test branch. It will be moved after merge

templates/Dockerfile-rpm.tmpl
```diff
@@ -1,6 +1,11 @@
 # vim:set ft=dockerfile:
 FROM %%os%%:%%release%%
 
+RUN [[ %%os%% != centos ]] || [[ %%release%% != 8 ]] || ( \
```
gurkanindibay: centos-8 default repository mirrors are not working so I changed the mirror into a working one

scripts/fetch_and_build_rpm
```diff
@@ -33,7 +33,7 @@ E_O_USAGE
 
 # sets the next version variable used during non-release builds
 setnextversion() {
-    baseversion=$(echo "$1" | tr '~' '-' | cut -d- -f1)
+    baseversion=$(echo "$1" | tr '~' '-' | tr '_' '-' | cut -d- -f1)
```
gurkanindibay, line 40: Sample version with beta is 102.4_beta. To be able to get the next version first we need to strip _beta (release stage). This line strips beta

scripts/fetch_and_build_rpm
```diff
@@ -58,6 +58,11 @@ nightlyref="${nightlyref:-master}"
 releasepg="${releasepg:-11,12,13}"
 nightlypg="${nightlypg:-${releasepg}}"
 versioning="${versioning:-simple}"
+if [[ "${pkglatest}" == *"beta"* ]]; then
```
gurkanindibay, line 65: This block defines beta prefix

scripts/fetch_and_build_rpm
```diff
@@ -166,8 +166,14 @@ case "${1}" in
     release)
         # add minor/major version to package name if using fancy versioning
         if [ "${versioning}" == 'fancy' ]; then
+
             infix=$(echo "${packageversion}" | grep -oE '^[0-9]+\.[0-9]+' | tr -d '.')
-            sed -i -E "1i %global pkginfix ${infix}" "${builddir}/${pkgname}.spec"
+            if [ "${release_type}" == "stable" ]; then
```
gurkanindibay, line 182: This block adds prefix for package

scripts/fetch_and_build_deb
```diff
@@ -161,9 +166,15 @@ case "${1}" in
     release)
         # add minor/major version to package name if using fancy versioning
         if [ "${versioning}" == 'fancy' ]; then
+            declare release_prefix=''
```
gurkanindibay, line 169: This block adds prefix  into package name

scripts/fetch_and_build_deb
```diff
@@ -200,3 +211,6 @@ for file in ../*.ddeb; do
 done
 ## Copy all deb packages out of docker image
 cp ../*.deb /packages
+
+cat debian/rules
+cat debian/control.in
```
onurctirtir: maybe echo something like below to let the people reviewing github action job:

echo "Printing debian/rules and debian/control.in files"

scripts/fetch_and_build_rpm
onurctirtir: maybe put a comment for that:

```
# First strip by foo, then strip by bar so that we can bla bla ..
# e.g.: For 102.4_beta, we will first remove "_beta" (if exists) then do bla bla
```

scripts/fetch_and_build_rpm
```diff
@@ -78,11 +83,6 @@ case "${1}" in
         packageversion=${pkglatest%-*}
         releasetag="v${packageversion%.citus}"
 
-        packagerelease=$(echo "${pkglatest#*-}" | sed -E 's/^[0-9.]+//')
-        if [ -n "${packagerelease}" ]; then
-            releasetag="v${packageversion}-${packagerelease}"
-        fi
```
onurctirtir: why are we removing this block ?

update_images_test
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
```
onurctirtir: What about refactoring the existing script into a common bash function and then call that function both from `update_images` & `update_images_test` by passing some params for beta/actual images ?

Also, maybe `update_images_test` -> `update_test_images`.

ci/push_images
```diff
@@ -12,21 +12,29 @@ badusage=64
...
 nprocs="${1:-1}"
 
-if [[ ${CURRENT_BRANCH} != "develop" ]] ; then
+if [[ ${TEST} == "false" && ${CURRENT_BRANCH} != "develop" ]] ; then
     echo "PR is not merged to develop, not pushing to DockerHub"
     exit 0;
 fi
 
+if [ "${TEST}" == "true" ]; then
+    image_name="packaging-test"
```
onurctirtir, line 21: what about "packaging-beta" ?

.github/workflows/build-package.yml
onurctirtir: Nice but only asking to better understand; why do we need that ?

onurctirtir (commented):
Overall LGTM, and thank you so much for detailed pr comments / pr description.

templates/Dockerfile-rpm.tmpl
onurctirtir: I guess you reverted this change later, right ?

scripts/fetch_and_build_deb
onurctirtir: I'm not able to see this change anymore, maybe a github bug.
Anyway, resolving this comment.

.github/workflows/build-package.yml
gurkanindibay: Actually it was a bug that I've recently discovered. push_images are working for all images. But I want each platform push its own image. So I fixed like that

hanefi:
We created the beta packages without merging this PR here. What would be different if we actually merged this one before creating the packages?

Would we have different package names?

Now that we created beta packages, is it safe to close this PR without merging?

ci/push_images
gurkanindibay: There is no need for seperate beta packages. seperate test packages are needed to be able to test packaging images before creating the actual ones

scripts/fetch_and_build_rpm
gurkanindibay: Added comment below
# Below two lines, strips patch version from the version.
    # First line replaces '~' and '_' characters and splits the version by '-' and gets the first
    # from the array.
    # Second line strips '.citus' suffix
    # e.g. input->11.0.0_beta-1.citus; output->11.0.0

templates/Dockerfile-rpm.tmpl
gurkanindibay: No, I fixed it in commit  below in parallel and merged so PR does not show the difference now
https://github.com/citusdata/packaging/commit/eb05a7a46583ce5cb117fe4da33a823a2fbde000

scripts/fetch_and_build_rpm
gurkanindibay: I found out that it has no use. But I'm adding it again with the comment includes my comments about its uselessness.

update_images_test
gurkanindibay: Removed it and performed it as you suggest

gurkanindibay:
> We created the beta packages without merging this PR here. What would be different if we actually merged this one before creating the packages?
> 
> Would we have different package names?
> 
> Now that we created beta packages, is it safe to close this PR without merging?

Accidentally I created the images without merging PR. But we need to merge it.

templates/Dockerfile-deb.tmpl
```diff
@@ -31,7 +31,7 @@ RUN set -ex; \
     export GNUPGHOME="$(mktemp -d)"; \
 # Fix ipv6 issue on travis: https://github.com/f-secure-foundry/usbarmory-debian-base_image/issues/9#issuecomment-466594168
     echo "disable-ipv6" >> $GNUPGHOME/dirmngr.conf; \
-    gpg --batch --keyserver keyserver.ubuntu.com --recv-keys "$key"; \
+    gpg --batch --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys "$key"; \
```
gurkanindibay, line 34: In local development without protocol and port, firewall could block downloading gpg key. With this addition. keys are being downloaded from 80 port which is open for nearly all firewalls

.github/workflows/build-package-test.yml
hanefi: This should be done in a separate PR IMO. You touched 47 files here and it is not easy to review as is.

.github/workflows/build-package.yml
```diff
@@ -36,14 +37,21 @@ jobs:
...
           - oraclelinux,7
-          - ubuntu,focal
-          - ubuntu,bionic
-          - pgxn
+        POSTGRES_VERSION:
+          - 10
+          - 11
+          - 12
+          - 13
+          - 14
```
hanefi, line 47: Shouldn't the PG values be inferred from the matrix?

.github/workflows/build-package.yml
```diff
@@ -36,14 +37,21 @@ jobs:
...
+        include:
+          - TARGET_PLATFORM: debian,bullseye
+          - TARGET_PLATFORM: debian,buster
+          - TARGET_PLATFORM: debian,stretch
+          - TARGET_PLATFORM: ubuntu,focal
+          - TARGET_PLATFORM: ubuntu,bionic
+          - TARGET_PLATFORM: pgxn
```
hanefi, line 54: What is the purpose of this include line? It is not so clear to me.

ci/push_images
hanefi: IMO, it makes more sense to have this comments in the code, not on Github UI.

hanefi:
FYI, You need to resolve conflicts here.

.github/workflows/build-package.yml
gurkanindibay: Actually no. We need to create each of the supported packaging images. Matrix is for to choose the postgres version releated to citus version. For example we may release 8.x 10.x and 11.x at the same time. For centos-8, we need pg 10 and 11 for 8.x , 12,13 and 14 for 10.x and 13 and 14 for 11.x. So we need all the images at the same time.

.github/workflows/build-package.yml
gurkanindibay: I needed to seperate centos builds for each pg version since I got errors for centos 8. Before this development, 4 postgres docker image was build at the same process but it was very error prone and failed frequently. Now I seperated them and if one of them fail I have an opportunity to rerun just one docker image
Therefore I created a cartesian multiplication of PLATFORM and POSTGRES_VERSION for rpm based images.
However cartesian multiplication does not work for deb based images since there is only one image for each platform
include part of yml adds processes one by one without cartesian multipl.
For reference: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations

scripts/fetch_and_build_rpm
gurkanindibay: I readded it @hanefi could we remove this block?

hanefi: I could not really come up with a meaningful package name for citus where it makes sense to keep this code block.

However I am not so sure about package names for other extension. Maybe it is best to keep it. I do not have any strong opinions on this matter.

hanefi (approved):

PR #895, opened 2022-03-12
Updates tools versions in pipelines

gurkanindibay:
Additionally, new warning appeared for debian and added them into ignore file since I could not find a proper solution

### DIFF
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -38,7 +38,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Clone tools branch
-        run: git clone -b v0.8.10 --depth=1  https://github.com/citusdata/tools.git tools
+        run: git clone -b v0.8.12 --depth=1  https://github.com/citusdata/tools.git tools
 
       - name: Install package dependencies
         run: sudo apt-get update && sudo apt-get install libcurl4-openssl-dev libssl-dev python3-testresources
--- a/packaging_ignore.yml
+++ b/packaging_ignore.yml
@@ -10,7 +10,6 @@ base:
   - ".* warning: overriding recipe for target 'check'"
   - "dpkg-genchanges: warning: the current version .* is earlier than the previous one .*"
   - ".*: warning: '__s' may be used uninitialized in this function \\[-Wmaybe-uninitialized\\].*"
-  - ".* unused substitution variable \\$\\{misc:Depends\\}.*"
   - ".*: substitution variable \\$\\{postgresql:Depends\\} unused, but is defined.*"
   - ".* unused substitution variable \\$\\{postgresql:Depends\\}.*"
 debian:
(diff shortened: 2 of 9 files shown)


packaging_ignore.yml
```diff
@@ -10,6 +10,9 @@ base:
   - ".* warning: overriding recipe for target 'check'"
   - "dpkg-genchanges: warning: the current version .* is earlier than the previous one .*"
   - ".*: warning: '__s' may be used uninitialized in this function \\[-Wmaybe-uninitialized\\].*"
+  - ".* unused substitution variable \\$\\{misc:Depends\\}.*"
```
gurkanindibay, line 13: new warnings appearead and I can not remove these ones with changing control.in file.

.github/workflows/build-package.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Clone tools branch
-        run: git clone -b v0.8.10 --depth=1  https://github.com/citusdata/tools.git tools
+        run: git clone -b v0.8.12 --depth=1  https://github.com/citusdata/tools.git tools
```
gurkanindibay: Will move the tag into develop branch after tools PR https://github.com/citusdata/tools/pull/216 merge

packaging_ignore.yml
onurctirtir: can't we just remove `${misc:Depends}` from `Depends:`, as in https://github.com/citusdata/packaging/commit/fd537ba0dd3736512a337a281fd9aaf52a6b862a ?

Btw, what does having or not having `${misc:Depends}` in `Depends :` line mean ? Is there any docs that you can share here so that I can better understand ?

onurctirtir (commented):
(left a comment above)

packaging_ignore.yml
gurkanindibay: You can see the result with misc:Depends removed from control.in
https://github.com/citusdata/packaging/actions/runs/1981648109
The document for this parameter 
https://www.debian.org/doc/manuals/maint-guide/dreq.en.html#control

hanefi: I am ok with this change. The `${misc:Depends} may be helpful in case `dh_*` scripts actually find some dependencies that should be added to the dependencies. We may observe some empty lists from time to time, but that's ok I guess.

hanefi (dismissed):

packaging_ignore.yml
```diff
@@ -13,7 +13,7 @@ base:
   - ".* unused substitution variable \\$\\{misc:Depends\\}.*"
   - ".*: substitution variable \\$\\{postgresql:Depends\\} unused, but is defined.*"
   - ".* unused substitution variable \\$\\{postgresql:Depends\\}.*"
-
+
```
hanefi, line 16:
```suggestion

```

hanefi:
Now that we do not need to update warning ignore rules, I do not see the point in updating all the packaging branches in this repo.

What other branches do you plan to update?

hanefi (approved):
one minor spacing issue

PR #899, opened 2022-03-24
Bump to citus 11.0.0_beta

gurkanindibay:


postgres-matrix.yml
```diff
@@ -12,3 +12,5 @@ version_matrix:
       postgres_versions: [12, 13]
   - 10.2:
       postgres_versions: [12, 13, 14]
+  - 11.0:
```
onurctirtir, line 15: really great to see that we have such a matrix :)

pkgvars
```diff
@@ -1,4 +1,4 @@
 pkgname=citus
 pkgdesc='Citus (Open-Source)'
-pkglatest=10.2.5.citus-1
+pkglatest=11.0.0_beta.citus-1
```
onurctirtir, line 3: wondering should this also be `11.0.0-beta` ?

onurctirtir: nvm, it shouldn't

gurkanindibay: No our tag is 11.0.0_beta as well

onurctirtir (approved):

PR #900, opened 2022-03-25
Bump pgxn to version 11.0.0_beta

gurkanindibay:


META.json
```diff
@@ -2,21 +2,21 @@
...
+   "version": "11.0.0-beta",
    "maintainer": "\"Citus Data\" <packaging@citusdata.com>",
    "license": "agpl_3",
    "provides": {
       "citus": {
          "abstract": "Citus Distributed Database",
          "file": "citus.so",
          "docfile": "README.md",
-         "version": "10.2.4"
+         "version": "11.0.0-beta"
```
gurkanindibay, line 13: pgxn standard does not accept _ so I need to change the '_' into -
Since we are using pkgvars pkglatest parameter to get the code from github, it will not affect us

META.json
```diff
@@ -25,7 +25,7 @@
          }
       }
    },
-   "release_status": "stable",
+   "release_status": "testing",
```
gurkanindibay, line 28: Defines beta package

hanefi (approved):

PR #909, opened 2022-04-26
Adds GH Actions pipeline, postgres-matrix and packaging_ignore file

gurkanindibay:


.github/workflows/build-package.yml
```diff
@@ -0,0 +1,61 @@
...
+
+jobs:
+  build_package:
+    name: Build package
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - el/8
```
hanefi, line 27: Do we actually use el/8 or el/7 ?

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,61 @@
...
+      - "**"
+
+  workflow_dispatch:
+
+jobs:
+  build_package:
+    name: Build package
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
hanefi, line 23: Why do we not fail fast? Any particular reason?

packaging_ignore.yml
```diff
@@ -0,0 +1,23 @@
...
+  - ".* warning: overriding recipe for target 'check'"
+  - "dpkg-genchanges: warning: the current version .* is earlier than the previous one .*"
+  - ".*: warning: '__s' may be used uninitialized in this function \\[-Wmaybe-uninitialized\\].*"
+  - ".* unused substitution variable \\$\\{misc:Depends\\}.*"
+  - ".*: substitution variable \\$\\{postgresql:Depends\\} unused, but is defined.*"
+  - ".* unused substitution variable \\$\\{postgresql:Depends\\}.*"
+
+debian:
+  - ".*: W: invalid-license Commercial"
+  - ".*: W: no-documentation"
```
hanefi, line 19: Do we want to keep Debian specific rules here as well?

I am ok with keeping those here for consistency. There is no harm in doing this I think.

.github/workflows/build-package.yml
```diff
@@ -0,0 +1,61 @@
+name: Build Package
```
hanefi: can we give this a better name that also contains azure gdpr in it?

.github/workflows/build-package.yml
gurkanindibay: el/7 also exist in travis file. So I added el/7 as well

.github/workflows/build-package.yml
gurkanindibay: Yes matrix jobs are independent from each other. So If a package job fails for an os version, I want other jobs continue. For some cases, there may be some os specific problems and I don't want these spesific problems to block other os jobs

packaging_ignore.yml
gurkanindibay: I just want it consistent as much as possible as you mentioned

.github/workflows/build-package.yml
gurkanindibay: Changed  and gave a more descriptive name

hanefi (approved):

PR #925, opened 2022-05-23
Bump to citus_enterprise 9.1.3

gurkanindibay:


hanefi (approved):

hanefi:
I suggest we remove the new ignore lines after all backports are finished.

We do not expect to see them in newer versions, and the tests should not pass if they start to show up for some reason.

gurkanindibay:
> I suggest we remove the new ignore lines after all backports are finished.
> 
> We do not expect to see them in newer versions, and the tests should not pass if they start to show up for some reason.

I will

> I suggest we remove the new ignore lines after all backports are finished.
> 
> We do not expect to see them in newer versions, and the tests should not pass if they start to show up for some reason.

https://github.com/citusdata/packaging/issues/926

PR #927, opened 2022-05-23
Bump to citus_enterprise 9.0.3

gurkanindibay:


citus-enterprise.spec
```diff
@@ -46,7 +46,7 @@ if [ "$(printf '%s\n' "$requiredgccver" "$currentgccver" | sort -V | head -n1)"
     exit 1
 fi
 
-%configure PG_CONFIG=%{pginstdir}/bin/pg_config --with-extra-version="%{?conf_extra_version}" CC=$(command -v gcc)
+%configure PG_CONFIG=%{pginstdir}/bin/pg_config --with-extra-version="%{?conf_extra_version}" --with-security-flags CC=$(command -v gcc)
```
hanefi, line 49: Why did we need to remove the flags for `v9.1.3` but not `v9.0.3` ?

That is rather strange to me. Maybe we should also fix that in `9.1` and create the packages for that version. `9.0` is fine though.

hanefi (approved):

citus-enterprise.spec
hanefi: It turns out that `--with-security-flags` is not supported on `9.1` as well as `9.0`.

It would have made more sense to not add that option to configure here. Apparently there is no harm in that as we were able to create the packages just fine.

PR #928, opened 2022-05-23
Bump to citus_enterprise 8.3.4

gurkanindibay:


citus-enterprise.spec
```diff
@@ -1,20 +1,19 @@
...
 Provides:	citus_%{pgmajorversion}
 Conflicts:	citus_%{pgmajorversion}
-Version:	9.0.3.citus
+Version:	8.3.4.citus
 Release:	1%{dist}
-License:	Commercial
+License:	AGPLv3
 Group:		Applications/Databases
-Source0:	https://github.com/citusdata/citus-enterprise/archive/v9.0.3.tar.gz
+Source0:       https://github.com/citusdata/citus-enterprise/archive/v8.3.4.tar.gz
```
hanefi:
```suggestion
Source0:	https://github.com/citusdata/citus-enterprise/archive/v8.3.4.tar.gz
```

citus-enterprise.spec
```diff
@@ -1,20 +1,19 @@
...
 
 Summary:	PostgreSQL-based distributed RDBMS
 Name:		%{sname}%{?pkginfix}_%{pgmajorversion}
 Provides:	citus_%{pgmajorversion}
 Conflicts:	citus_%{pgmajorversion}
-Version:	9.0.3.citus
+Version:	8.3.4.citus
 Release:	1%{dist}
-License:	Commercial
+License:	AGPLv3
```
hanefi: I do not think `AGPLv3` is a wise choice here.

Let's keep it as `Commercial` please.

```suggestion
License:	Commercial
```

citus-enterprise.spec
```diff
@@ -1,20 +1,19 @@
-%global pgmajorversion 11
-%global pgpackageversion 11
+%global pgmajorversion 10
+%global pgpackageversion 10
```
hanefi, line 2: Not really necessary as these values should be updated by `citus_dev` eventually.

hanefi (dismissed):
Approving with 2 suggestions that should not affect the builds, but only the metadata of the generated packages and proper spec file format

PR #929, opened 2022-05-24
Rollbacks build changes for out of order builds

gurkanindibay:


hanefi (approved):
I am ok with all the changes here as long as we will not push new packages to our repositories.

PR #943, opened 2022-07-13
Bump to citus 11.0.4

gurkanindibay:


hanefi (approved):
I guess it is time to update the templates that generate commits that bump package versions.

Please use squash strategy when merging

PR #945, opened 2022-07-25
Adds changes for one branch pipelines

gurkanindibay:
Changes drop into three categories
1. One branch pipelines needs images to stay in working state after image start. With the new environment variable which is set in one branch pipeline, image stay in working state
2. Images are now being pushed into new Azure Container registry, 'citusengine' which is used by OneBeanch pipeline
3. sudo package is added into deb based docker files since it is required for OneBranch and does not exist in debian based images by default

Note. Since nightlies are failing nightly tests in this PR fails as well. So I could not fix the problem in here I will fix it in another PR. Therefore please ignore nightly pipeline test fail for this PR
Related PRS: 
https://github.com/citusdata/tools/pull/233
https://msdata.visualstudio.com/Database%20Systems/_git/citus-packaging/pullrequest/817728


.github/workflows/build-package-test.yml
```diff
@@ -73,7 +75,7 @@ jobs:
           POSTGRES_VERSION: ${{ matrix.POSTGRES_VERSION }}
 
       - name: Clone tools repo for test
-        run: git clone -b v0.8.15 --depth=1  https://github.com/citusdata/tools.git tools
+        run: git clone -b one_branch_additions --depth=1  https://github.com/citusdata/tools.git tools
```
gurkanindibay: TODO: Branch name should be replaced with the new tools tag after tools PR merge

ci/push_images
```diff
@@ -64,4 +64,10 @@ done <"${topdir}/os-list.csv"
 
 echo -e "${args}" | xargs -t -L1 -P "${nprocs}" docker
 
+docker tag citus/"${image_name}":"${tag}" citusengine.azurecr.io/citus/"${image_name}":"${tag}"
```
gurkanindibay, line 67: ACR Push code block

dockerfiles/centos-7-pg10/scripts/fetch_and_build_rpm
```diff
@@ -1,7 +1,16 @@
 #!/bin/bash
 
 # make bash behave
-set -euo pipefail
+set -eo pipefail
```
gurkanindibay: When initializing the docker image in OneBranch, docker run process fall into this block so it does not die

scripts/fetch_and_build_rpm
```diff
@@ -150,6 +159,8 @@ mkdir -p "${rpmbuilddir}"
 pkgsrcdir="${builddir}/${pkgname}-${packageversion}"
 mkdir "${pkgsrcdir}"
 
+
+
```
onurctirtir: nit: maybe not add those two unnecessary lines

scripts/fetch_and_build_rpm
```diff
@@ -1,7 +1,16 @@
 #!/bin/bash
 
 # make bash behave
-set -euo pipefail
+set -eo pipefail
+# In one branch we execute commands inside One Branch steps, since One Branch does not allow executing docker inside
+# docker. Additionally, Onebranch needs containers not to close so we make it hang for OneBranch to be able to
+# execute commands.
+if [[ -z ${CONTAINER_BUILD_RUN_ENABLED} ]]; then
```
onurctirtir: we don't have to drop set -u here, the following would help:

```suggestion
set -euo pipefail
# In one branch we execute commands inside One Branch steps, since One Branch does not allow executing docker inside
# docker. Additionally, Onebranch needs containers not to close so we make it hang for OneBranch to be able to
# execute commands.
if [ "${CONTAINER_BUILD_RUN_ENABLED:-""}" == "" ]; then
```

onurctirtir (approved):
please consider the comments before merging

.github/workflows/build-package-test.yml
hanefi: need to update after merging tools PR

.github/workflows/build-package.yml
```diff
@@ -74,7 +76,8 @@ jobs:
           POSTGRES_VERSION: ${{ matrix.POSTGRES_VERSION }}
 
       - name: Clone tools repo for test
-        run: git clone -b v0.8.15 --depth=1  https://github.com/citusdata/tools.git tools
+        run: git clone -b one_branch_additions --depth=1  https://github.com/citusdata/tools.git tools
+
```
hanefi: why the extra line here?

```suggestion
```

scripts/fetch_and_build_rpm
hanefi: +1 to this idea. Nice suggestion too.

templates/Dockerfile-deb.tmpl
```diff
@@ -71,6 +71,7 @@ RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ %%release%%-pgdg main' >
         liblz4-1 \
         libzstd1 \
         libzstd-dev \
+        sudo \
```
hanefi, line 74: It is not usually advised to use sudo in docker environment. Why is this addition necessary?

Did you consider `gosu` or `USER` docker directives instead?

scripts/fetch_and_build_rpm
```diff
@@ -1,7 +1,16 @@
...
+  echo "INFO: Image working in waiting mode. Not executing build script"
+  tail -f /dev/null
+  exit 0
```
hanefi: If this is an error, does it make sense to return 0?

hanefi: Also, is this even reachable? How do we terminate or stop this docker image?

gurkanindibay: You're right I added it just to show that it is not error and a valid code path. It's better to remove it. It may be confusing

templates/Dockerfile-deb.tmpl
gurkanindibay: Actually OneBranch needs it. They use sudo 
https://msdata.visualstudio.com/Database%20Systems/_build/results?buildId=68259209&view=logs&j=570eab73-0366-550a-dd5d-1b82c1121f48&t=49b1901d-5c98-4bfe-afc8-a8deb533389f&l=98

.github/workflows/build-package-test.yml
gurkanindibay: Fixed

gurkanindibay: Fixed

templates/Dockerfile-deb.tmpl
hanefi: I guess it is ok then. We can not do anything about this if OneBranch decides to use an antipattern 😮‍💨

scripts/fetch_and_build_rpm
hanefi: bump

scripts/fetch_and_build_rpm
hanefi: How do we terminate the docker container then? Does it keep running forever indefinitely?

scripts/fetch_and_build_rpm
hanefi: bump

hanefi (approved):

scripts/fetch_and_build_rpm
gurkanindibay: docker stop :)

PR #947, opened 2022-07-26
Fixes nightly problems for rpm packages because of new columnar files

gurkanindibay:


citus.spec
```diff
@@ -60,6 +60,19 @@ echo %{pginstdir}/lib/%{sname}.so >> installation_files.list
...
+if [ ${#columnar_sql_files[@]} -gt 0 ]; then
+    echo %{pginstdir}/share/extension/columnar-*.sql >> installation_files.list
+fi
+
+citus_columnar_sql_files=(`find %{buildroot}%{pginstdir}/share/extension -maxdepth 1 -name "citus_columnar-*.sql"`)
+if [ ${#citus_columnar_sql_files[@]} -gt 0 ]; then
+    echo %{pginstdir}/share/extension/citus_columnar-*.sql >> installation_files.list
+fi
+
+[[ -f %{pginstdir}/share/extension/citus_columnar.control ]] && echo %{pginstdir}/share/extension/citus_columnar.control >> installation_files.list
```
onurctirtir: why do we have two lines for citus_columnar.control L64 & L75

onurctirtir (commented):
And definitely need to locally-build & test el7 package, e.g.: whether we can install it and then can do `create extension citus;` via psql.

citus.spec
gurkanindibay: my fault. Removed it.

gurkanindibay:
> And definitely need to locally-build & test el7 package, e.g.: whether we can install it and then can do `create extension citus;` via psql.

Tested successfully

onurctirtir (approved):

hanefi (approved):

Issue #948, opened 2022-07-28
Consider storing tools version in env


hanefi:
We have many workflows that want to use the latest version of tools repository. This means that we change many workflow definitions on seemingly irrelevant PRs such as https://github.com/citusdata/packaging/pull/945

I suggest we come up with a way to share the common tools version that is used in all the workflows in a single place. A possible solution is to create an env file. See https://arinco.com.au/blog/github-actions-share-environment-variables-across-workflows/

gurkanindibay:
It is not correct to set all the pipelines equal. In this case, we don't need to version. We can use the latest version in this case.

gurkanindibay:
I'm updating when it is needed after I execute tests one by one

PR #949, opened 2022-07-29
Updates tools version for community packaging

gurkanindibay:


hanefi (approved):
Please consider implementing https://github.com/citusdata/packaging/issues/948 so that it is easier and cleaner to make similar changes.

PR #953, opened 2022-08-02
Adds pg 15 beta support in packaging images

gurkanindibay:
Postgres 15 is in beta phase. Wİth this PR, packaging images will be changed to build postgres packages. All the files under dockerfiles is autogenerated using update_dockerfiles so no need to review.


.github/workflows/build-package.yml
```diff
@@ -55,6 +56,9 @@ jobs:
           - TARGET_PLATFORM: ubuntu,bionic
           - TARGET_PLATFORM: ubuntu,jammy
           - TARGET_PLATFORM: pgxn
+        exclude:
+          - TARGET_PLATFORM: oraclelinux,7
+            POSTGRES_VERSION: 15
```
gurkanindibay, line 61: oracle linux 7 getting error while downloading postgres 15 as below
https://github.com/citusdata/packaging/runs/7646275178?check_suite_focus=true#step:5:776
So excluded pg 15 for oracle linux 7

supported_versions_pg15.patch
```diff
@@ -0,0 +1,10 @@
+--- supported-versions	2022-08-02 12:11:38.188738000 +0000
++++ supported-versions_with_15	2022-08-02 12:14:31.608463100 +0000
+@@ -137,6 +137,7 @@
+ 	12
+ 	13
+ 	14
++	15
```
gurkanindibay: pg 15 support should be added to build successfully

templates/Dockerfile-deb.tmpl
```diff
@@ -88,6 +89,10 @@ RUN curl -s https://packagecloud.io/install/repositories/citusdata/community/scr
 # patch pg_buildext to use multiple processors
 COPY make_pg_buildext_parallel.patch /
 RUN patch `which pg_buildext` < /make_pg_buildext_parallel.patch
+# Added for pg15 beta package support. Without this pg_buildext gives below error
+# No current PostgreSQL versions are supported by this package
+COPY supported_versions_pg15.patch /
+RUN patch /usr/share/postgresql-common/supported-versions < /supported_versions_pg15.patch
```
gurkanindibay: patch is applied to be able to add support for pg build on debian

gurkanindibay:
@hanefi when could you check the PR? After completion of this PR, packaging images will be created and there are some changes for all-citus branch

ci/push_images
```diff
@@ -62,6 +62,8 @@ while read -r line; do
     fi
 done <"${topdir}/os-list.csv"
 
+echo "Arguments: ${args}"
+
```
hanefi: I suggest we remove these lines before merge to make sure we do not leak any secrets.

templates/Dockerfile-rpm.tmpl
```diff
@@ -41,9 +41,9 @@ RUN if [[ %%os%%   == oraclelinux ]] && [[ %%release%% == 7 ]]; then yum install
         && rpm -ivh lz4-devel-r131-1.el6.x86_64.rpm;  \
         else yum install -y lz4 lz4-devel; fi
 
-
 # install build tools and PostgreSQL development files
-RUN ( yum install -y https://%%rpm_url%% \
+RUN ( yum install -y https://%%rpm_url%%  \
+    && ( [[ %%pgshort%% != 15 ]] || sed -i '/\[pgdg15-updates-testing\]/{n;n;n;s/.*/enabled=1/}' /etc/yum.repos.d/pgdg-redhat-all.repo ) \
```
hanefi, line 46: I find it hard to decrypt this sed script. Can we use the cli instead?

```
yum-config manager --enable pgdg15-updates
```

supported_versions_pg15.patch
hanefi: This patch is bound to fail once postgresql-client-common package is updated after PG15 is released.

I suggest you use some of the following alternatives to override this value. See https://salsa.debian.org/postgresql/postgresql-common/-/blob/master/debian/supported-versions#L21-26

```
# * PG_SUPPORTED_VERSIONS
# * DEB_PG_SUPPORTED_VERSIONS
# * ~/.pg_supported_versions
# * /etc/postgresql-common/supported_versions
# (in that order) can be used to override the defaults. (Tokens separated by
# newlines.)
```

hanefi (changes requested):
I suggest you consider using environment variables or a configuration file to override the list of supported PostgreSQL versions. The current solution is not going to age nicely.

supported_versions_pg15.patch
gurkanindibay: After beta release process ends, we already need to rollback the changes we've made. So we can rollback this one as well

templates/Dockerfile-rpm.tmpl
gurkanindibay: good catch  New day new knowledge :)

gurkanindibay:
> I suggest you consider using environment variables or a configuration file to override the list of supported PostgreSQL versions. The current solution is not going to age nicely.

Good catch. I used DEB_PG_SUPPORTED_VERSIONS instead. Thanks

templates/Dockerfile-rpm.tmpl
gurkanindibay: I tested it but it did not worked for centos 8.  Rolled back
https://github.com/citusdata/packaging/runs/7667950735?check_suite_focus=true#step:5:2894

gurkanindibay: BTW I need to replace it after Postgres GA. SO there is no harm to keep  it temporarily I think

hanefi: I am cool with the current way as long as it works.

rhel 8 uses dnf instead of yum. There may be some alternative command that we need on that distro. No need to spend more time on this I guess.

templates/Dockerfile-deb.tmpl
```diff
@@ -89,8 +90,12 @@ RUN curl -s https://packagecloud.io/install/repositories/citusdata/community/scr
 COPY make_pg_buildext_parallel.patch /
 RUN patch `which pg_buildext` < /make_pg_buildext_parallel.patch
 
+# Added for pg15 beta package support. Without this pg_buildext gives below error
```
hanefi: I do not see any error message here. I guess you can remove part of this comment as it is quite self explanatory already.

```suggestion
# Added for pg15 beta package support.
```

templates/Dockerfile-deb.tmpl
```diff
@@ -89,8 +90,12 @@ RUN curl -s https://packagecloud.io/install/repositories/citusdata/community/scr
...
+ENV DEB_PG_SUPPORTED_VERSIONS="10 11 12 13 14 15"
+
 # place scripts on path and declare output volume
 ENV PATH /scripts:$PATH
+
```
hanefi:
```suggestion
```

hanefi (approved):

ci/push_images
gurkanindibay: Removed

PR #956, opened 2022-08-05
Adds pg 15 beta support

gurkanindibay:
This branch is a temporary branch to be able to handle pg 15 temporary changes especially  --without-pg-version-check flag
Since it includes temporary flag, it will not be merged into all-citus

### DIFF
--- a/.github/workflows/build-citus-community-nightlies.yml
+++ b/.github/workflows/build-citus-community-nightlies.yml
@@ -25,7 +25,8 @@ jobs:
         platform:
           - el/7
           - el/8
-          - ol/7
+          # temporarily removed from nightlies since pg15 beta is not being downloaded for ol/7
+          # - ol/7
           - debian/stretch
           - debian/buster
           - debian/bullseye
--- a/.github/workflows/build-citus-community-nightlies.yml
+++ b/.github/workflows/build-citus-community-nightlies.yml
@@ -45,7 +45,7 @@ jobs:
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
 
       - name: Clone tools branch
-        run: git clone -b v0.8.17 --depth=1  https://github.com/citusdata/tools.git tools
+        run: git clone -b pg15_nightlies_exception --depth=1  https://github.com/citusdata/tools.git tools
 
       - name: Install package dependencies
         run: sudo apt-get update && sudo apt-get install libcurl4-openssl-dev libssl-dev python3-testresources
--- a/.github/workflows/build-citus-community-nightlies.yml
+++ b/.github/workflows/build-citus-community-nightlies.yml
@@ -25,8 +25,7 @@ jobs:
         platform:
           - el/7
           - el/8
-          # temporarily removed from nightlies since pg15 beta is not being downloaded for ol/7
-          # - ol/7
+          - ol/7
           - debian/stretch
           - debian/buster
           - debian/bullseye
--- a/citus.spec
+++ b/citus.spec
@@ -44,7 +44,7 @@ if [ "$(printf '%s\n' "$requiredgccver" "$currentgccver" | sort -V | head -n1)" 
     exit 1
 fi
 
-%configure PG_CONFIG=%{pginstdir}/bin/pg_config --with-extra-version="%{?conf_extra_version}" --with-security-flags CC=$(command -v gcc) --without-pg-version-check
+%configure PG_CONFIG=%{pginstdir}/bin/pg_config --with-extra-version="%{?conf_extra_version}" --with-security-flags CC=$(command -v gcc)
 make %{?_smp_mflags}
 
 %install
--- a/citus.spec
+++ b/citus.spec
@@ -44,7 +44,7 @@ if [ "$(printf '%s\n' "$requiredgccver" "$currentgccver" | sort -V | head -n1)" 
     exit 1
 fi
 
-%configure PG_CONFIG=%{pginstdir}/bin/pg_config --with-extra-version="%{?conf_extra_version}" --with-security-flags CC=$(command -v gcc)
+%configure PG_CONFIG=%{pginstdir}/bin/pg_config --with-extra-version="%{?conf_extra_version}" --with-security-flags CC=$(command -v gcc) --without-pg-version-check
 make %{?_smp_mflags}
 
 %install
--- a/debian/rules
+++ b/debian/rules
@@ -15,7 +15,7 @@ override_dh_auto_test:
 	# nothing to do here, see debian/tests/* instead
 
 override_dh_auto_configure:
-	+pg_buildext configure build-%v --with-extra-version="$${CONF_EXTRA_VERSION:-}" --with-security-flags
+	+pg_buildext configure build-%v --without-pg-version-check --with-extra-version="$${CONF_EXTRA_VERSION:-}" --with-security-flags
 
 override_dh_auto_install:
 	+pg_buildext install build-%v postgresql-%v-citus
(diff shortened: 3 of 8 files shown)


packaging_ignore.yml
```diff
@@ -21,6 +21,7 @@ base:
   - ".*: warning: 'RSA_generate_key_ex' is deprecated \\[-Wdeprecated-declarations\\]"
   - ".*: warning: 'EVP_PKEY_assign' is deprecated.*"
   - "\\d+ warnings generated."
+  - "configure: WARNING: unrecognized options: --without-pg-version-check"
```
hanefi, line 23: Why do we need this line? We added support for this in https://github.com/citusdata/citus/pull/6072 that is now in `main` branch.

If this option is not recognized, we will fail to build Citus on PG15 and we will have errors in later steps of our builds.

.github/workflows/build-citus-community-nightlies.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
 
       - name: Clone tools branch
-        run: git clone -b v0.8.17 --depth=1  https://github.com/citusdata/tools.git tools
+        run: git clone -b pg15_nightlies_exception --depth=1  https://github.com/citusdata/tools.git tools
```
hanefi, line 48: I guess this will be reverted along with all the other occurrences of the same branch.

postgres-matrix.yml
```diff
@@ -3,14 +3,14 @@ project_name: citus
...
+      postgres_versions: [ 11, 12, 13 ]
   - 10.1:
-      postgres_versions: [12, 13]
+      postgres_versions: [ 12, 13 ]
   - 10.2:
-      postgres_versions: [12, 13, 14]
+      postgres_versions: [ 12, 13, 14 ]
   - 11.0:
-      postgres_versions: [13,14]
+      postgres_versions: [ 13, 14, 15 ]
```
hanefi, line 16: I believe 11.1 will be the first version to support PG15.

pg_exclude.yml
```diff
@@ -0,0 +1,5 @@
+exclude:
+  nightly:
+    ol/7: [15]
+  release:
+    all: [15]
```
hanefi, line 5: How about moving this file and merging it with some of the current files that we already have? See https://github.com/citusdata/tools/pull/238#discussion_r941533392

packaging_ignore.yml
gurkanindibay: We are adding this warning temporarily. @onderkalaci told me to use this flag in the builds. You can see that I added that flag into both deb rule file and rpm spec file to be able to build pg 15.

.github/workflows/build-citus-community-nightlies.yml
gurkanindibay: After merge of PR https://github.com/citusdata/tools/pull/238 , I will put the latest tag

postgres-matrix.yml
gurkanindibay: I added this for nightlies. After 11.1 release I will change it. This is for beta process

pg_exclude.yml
gurkanindibay: Actually we will not need this file quite common so I prefer not adding this option into postgres_matrix file. Additionally, postgres matrix includes all versions which is some sort of a citus postgres version support matrix file. This file will be very long and adding additional things will make the file more difficult to read.
However, it is not a bad idea. If it is needed, I need to open a PR about that and merge them in the future since there are more prioritized tasks waiting my attention in my work queue.

citus.spec
```diff
@@ -44,7 +44,7 @@ if [ "$(printf '%s\n' "$requiredgccver" "$currentgccver" | sort -V | head -n1)"
     exit 1
 fi
 
-%configure PG_CONFIG=%{pginstdir}/bin/pg_config --with-extra-version="%{?conf_extra_version}" --with-security-flags CC=$(command -v gcc)
+%configure PG_CONFIG=%{pginstdir}/bin/pg_config --with-extra-version="%{?conf_extra_version}" --with-security-flags CC=$(command -v gcc) --without-pg-version-check
```
gurkanindibay: I added this check with the direction of @onderkalaci to be able to build nightlies for pg15

debian/rules
```diff
@@ -15,7 +15,7 @@ override_dh_auto_test:
 	# nothing to do here, see debian/tests/* instead
 
 override_dh_auto_configure:
-	+pg_buildext configure build-%v --with-extra-version="$${CONF_EXTRA_VERSION:-}" --with-security-flags
+	+pg_buildext configure build-%v --without-pg-version-check --with-extra-version="$${CONF_EXTRA_VERSION:-}" --with-security-flags
```
gurkanindibay: I added this check with the direction of @onderkalaci to be able to build nightlies for pg15

postgres-matrix.yml
hanefi: We use `11.1devel` version on the main branch.

1. Why do you need to update entries for `11.0` for nightlies? Should not that be 11.1?
2. Would not this change cause problems if we want to release a patch version for 11.0 that does not and will not support PG15.

I suggest we add 11.1 here and make sure nightlies actually use the values for that one.

packaging_ignore.yml
hanefi: We already have support for this option. If we observe such a warning there is something wrong with our builds. I agree that we should update debian/rules and citus.spec files. However I do not think that we should ever allow configure to throw this warning.

pg_exclude.yml
hanefi: I think that it is not too much of a hassle to concatenate the two files. I do not think that you need to update the structure.

How about the following suggestion for `postgres-matrix.yml`:

```yml
name: Postgres Version Matrix
project_name: citus

version_matrix:
  - 8.0:
      postgres_versions: [ 10, 11 ]
  - 9.0:
      postgres_versions: [ 11, 12 ]
  - 9.5:
      postgres_versions: [ 11, 12, 13 ]
  - 10.1:
      postgres_versions: [ 12, 13 ]
  - 10.2:
      postgres_versions: [ 12, 13, 14 ]
  - 11.0:
      postgres_versions: [ 13, 14 ]
  - 11.1:
      postgres_versions: [ 13, 14, 15 ]

exclude:
  nightly:
    ol/7: [ 15 ]
  release:
    all: [ 15 ]
```

postgres-matrix.yml
gurkanindibay: Current structure does not work like that. Nightlies are being processed seperately and evaluated with the current branch. Actually we do not make a version check since it is on main branch.

packaging_ignore.yml
gurkanindibay: I actualy does not understand what you mean. We do not throw an error specifically for this warning. We have a warning detector machanism which throws error other than items in packaging-ignore file. Therefore since this warning appears in the logs because we add this flag into our configuration, without adding this warning into ignore file, warning detector throws error.

pg_exclude.yml
gurkanindibay: Yes it can be like that but I wanted to keep postgres-matrix.yml for only versions. 
I opened an issue about that. I could start this issue after my prioritized tasks
https://github.com/citusdata/packaging/issues/958

gurkanindibay:
@hanefi I addressed all reviews. Could you check the PR again?

hanefi (dismissed):
I do not like the current design, but I will approve this in the hopes that we fix it soon. The main problem is captured in https://github.com/citusdata/packaging/issues/958

PR #957, opened 2022-08-05
Enable nightlies for pg15

gurkanindibay:
Removed ol/7 since pg 15 could not be downloaded for ol/7. I'm working another PR to be able to exclude pg versions that is not being supported by os release pair. With this PR  ol/7 nightlies could be baked for pg versions other than pg15
Note: With the approval of @metdos I removed both pgaf community and enterprise nightlies. Sİnce citus enterprise is not being used I removed it as well


.github/workflows/build-citus-community-nightlies.yml
```diff
@@ -1,7 +1,7 @@
 name: Build and publish citus community nightly packages
 
 env:
-  MAIN_BRANCH: "all-citus"
+  MAIN_BRANCH: "all-citus-pg15-beta"
```
hanefi, line 4: Do we intend to keep this change, or is it temporary?

.github/workflows/build-citus-community-nightlies.yml
```diff
@@ -27,7 +27,8 @@ jobs:
         platform:
           - el/7
           - el/8
-          - ol/7
+          # temporarily removed from nightlies since pg15 beta is not being downloaded for ol/7
+          # - ol/7
```
hanefi, line 31: Do we not create packages for all PG versions? If only PG15 does not work for ol/7 is there a way to disable only this combination?

hanefi (changes requested):
Now that we want to remove the workflow for enterprise, how about removing the word community from the single workflow we have left for Citus?

.github/workflows/build-citus-community-nightlies.yml
gurkanindibay: This is temporary. I will not merge the PR for all-citus-pg15-beta either. (https://github.com/citusdata/packaging/pull/956)

.github/workflows/build-citus-community-nightlies.yml
gurkanindibay: There is no way for it now. I opened the PR in tools to be able to handle it
https://github.com/citusdata/tools/pull/238

gurkanindibay: This PR is to enable nightlies without waiting the merge od this enablement PR. After merge of this PR there will be no need to disable ol/7 . We will just exclude pg15

gurkanindibay:
> Now that we want to remove the workflow for enterprise, how about removing the word community from the single workflow we have left for Citus?

I think we need more time for that. I've just removed community. Marlin still uses citus enterprise < 11. There may be some backports in the future for citus enterprise so it is too soon to make this changes

hanefi (approved):
Dont forget to revert back to all-citus branch before merging

gurkanindibay:
> Dont forget to revert back to all-citus branch before merging

No need to. We need pg 15 nightlies so all-citus-pg15-beta. I will revert back after citus will be ready without '--without-pg-version-check' flag

PR #959, opened 2022-08-10
Fixes package name problem and locale problem

gurkanindibay:
Currently, naming strucuture for rpm is as below
citus_13-11.1.0.citus-0.0.git.20220731.5490c85.el7.x86_64.rpm

When you try to install the package as instructed in citusdata download page, it doesn't work since package name does not include version just after citus as below
citus110_13-11.0.5.citus-1.el7.x86_64.rpm

With the new development, package name will be like below
citus110_13-11.1.0.citus-0.0.git.20220731.5490c85.el7.x86_64.rpm

Additionally I found a bug for centos 7 packages related to locale set. Since C.utf8 is not among supported locales for centos 7, the error below appears in the logs
"sh: warning: setlocale: LC_ALL: cannot change locale (C.utf8): No such file or directory"
When I check the commit history, I found out that the intention to add C.utf8 for Centos is just for el/8 , Not for el/7
Commit hash for this change: e864e961d385190786a5db74ac2e5dfc1132d167


scripts/fetch_and_build_rpm
```diff
@@ -173,38 +173,42 @@ sed -i -E -e "\\|^Source0:|s|https:.*|${tarballurl}|" \
     "${builddir}/${pkgname}.spec"
 
 osname=$(awk '{print $1}' /etc/system-release)
+os_release=$(awk '{print $4}' /etc/system-release)
 if [ "${osname}" == 'Oracle' ]; then
     locale='C'
-elif [ "${osname}" == 'Fedora' ] || [ "${osname}" == 'CentOS' ]; then
+elif [ "${osname}" == 'Fedora' ] || [[ "${osname}" == 'CentOS'  && "${os_release}" == 8* ]]; then
```
hanefi: I guess it is time that we remove all references to Fedora in our scripts. This hurts readability, and there are a lot of leftover code that do not really have any value.

Can we create an issue and track this please?

scripts/fetch_and_build_rpm
```diff
@@ -173,38 +173,42 @@ sed -i -E -e "\\|^Source0:|s|https:.*|${tarballurl}|" \
...
-            fi
-            sed -i -E "1i %global pkginfix ${package_prefix}" "${builddir}/${pkgname}.spec"
-        fi
         ;;
     nightly)
         msg="Nightly package. Built from ${nightlyref} "
         msg+=$(date +'on %l:%M %p (%Z) on %A, %B %Y' | tr -s ' ')
-        LC_ALL=${locale} rpmdev-bumpspec -c "${msg}" "${builddir}/${pkgname}.spec"
+        LC_ALL=${locale}
+        rpmdev-bumpspec -c "${msg}" "${builddir}/${pkgname}.spec"
```
hanefi: I think these two statements are meant to be on the same line.

```suggestion
        LC_ALL=${locale} rpmdev-bumpspec -c "${msg}" "${builddir}/${pkgname}.spec"
```

scripts/fetch_and_build_rpm
```diff
@@ -173,38 +173,42 @@ sed -i -E -e "\\|^Source0:|s|https:.*|${tarballurl}|" \
...
         sed -i -E 's/0.1.git/0.0.git/' "${builddir}/${pkgname}.spec"
+        sed -i -E "1i %global pkginfix ${package_prefix}" "${builddir}/${pkgname}.spec"
         ;;
     *)
         msg="Custom package. Built from ${gitsha:0:7} "
         msg+=$(date +'on %l:%M %p (%Z) on %A, %B %Y' | tr -s ' ')
-        LC_ALL=${locale} rpmdev-bumpspec -c "${msg}" "${builddir}/${pkgname}.spec"
+        LC_ALL=${locale}
+        rpmdev-bumpspec -c "${msg}" "${builddir}/${pkgname}.spec"
```
hanefi: I think these two statements are meant to be on the same line.

```suggestion
        LC_ALL=${locale} rpmdev-bumpspec -c "${msg}" "${builddir}/${pkgname}.spec"
```

scripts/fetch_and_build_rpm
```diff
@@ -173,38 +173,42 @@ sed -i -E -e "\\|^Source0:|s|https:.*|${tarballurl}|" \
     "${builddir}/${pkgname}.spec"
 
 osname=$(awk '{print $1}' /etc/system-release)
+os_release=$(awk '{print $4}' /etc/system-release)
```
hanefi, line 176: Nit, can we rename this to `osrelease` so we are consistent with the naming of `osname` ?

hanefi (commented):
Looks good overall. Will need to verify that the naming is still as expected after the comments are addressed.

hanefi:
Can we also trigger some nightly and regular image builds so that we can see everything works as expected?

scripts/fetch_and_build_rpm
gurkanindibay: https://github.com/citusdata/packaging/issues/960

scripts/fetch_and_build_rpm
gurkanindibay: You're right fixed as you suggested

scripts/fetch_and_build_rpm
gurkanindibay: You're right fixed as you suggested

scripts/fetch_and_build_rpm
gurkanindibay: Changed osname into os_name

gurkanindibay:
> Can we also trigger some nightly and regular image builds so that we can see everything works as expected?

Images should be created after merge of the PR to be able to see nightlies with the new images
I performed local test

gurkanindibay:
> > Can we also trigger some nightly and regular image builds so that we can see everything works as expected?
> 
> Images should be created after merge of the PR to be able to see nightlies with the new images I performed local test

image

scripts/fetch_and_build_rpm
```diff
@@ -172,34 +172,36 @@ sed -i -E -e "\\|^Source0:|s|https:.*|${tarballurl}|" \
...
+      package_prefix="${infix}"
+  else
+      package_prefix="${infix}_${release_type}"
+  fi
+  sed -i -E "1i %global pkginfix ${package_prefix}" "${builddir}/${pkgname}.spec"
+fi
+
 case "${1}" in
     release)
         # add minor/major version to package name if using fancy versioning
```
hanefi: This comment no longer makes sense. Can we update that with a simple explanation on why we do not have any steps for release?

hanefi (approved):
All good after we update a stale comment

PR #962, opened 2022-08-18
Updates el images to be able  to build pgazure packages

gurkanindibay:
Azure SDK is included in the pgazure source. They need to interact with azure and they couldn't find a proper os package. Therefore they need to build azure sdk from the source. cmake is required so I added cmake and its dependencies.
Changes are for el/7 and el/8 for now. I will add other os/release pairs in the future. Since Marlin is using el/7 we prioritized it first


templates/Dockerfile-rpm.tmpl
```diff
@@ -93,6 +93,20 @@ RUN { \
...
+RUN yum -y install  perl-IPC-Cmd libuuid-devel cmake3
+
+# by default git 1.8.x is being installed in centos 7
+# Git 2.7.1 is minimum requirement for cmake so we remove and reinstall latests git from an up-to-date repo
+# Symbolic link is not being created for cmake from cmake3 by default. Therefore, we need to create the link as well.
+# devtoolset-8-gcc-c++ is required to compile azure sdk in pg azure storage project
+RUN [[ %%os%% != centos ]] || [[ %%release%% != 7 ]] || ( \
+    yum -y install  perl-IPC-Cmd libuuid-devel cmake3 && \
+    ln -s /usr/bin/cmake3 /usr/bin/cmake  && \
+    yum -y remove git && \
```
onurctirtir, line 106: why we're removing "git" and then re-installing it ?

onurctirtir (approved):

templates/Dockerfile-rpm.tmpl
```diff
@@ -93,6 +93,20 @@ RUN { \
     } \
     && yum clean all
 
+# install cmake, devtoolset and its dependencies to build azure sdk
+RUN yum -y install  perl-IPC-Cmd libuuid-devel cmake3
```
hanefi, line 97: It is a good idea to run `yum clean all` to ensure we have a small docker image, and small docker cache usage for intermediate steps.

templates/Dockerfile-rpm.tmpl
hanefi, line 106: You do not need to remove git. `yum install git` will actually upgrade the package if it is already installed.

templates/Dockerfile-rpm.tmpl
```diff
@@ -93,6 +93,20 @@ RUN { \
...
+
+# by default git 1.8.x is being installed in centos 7
+# Git 2.7.1 is minimum requirement for cmake so we remove and reinstall latests git from an up-to-date repo
+# Symbolic link is not being created for cmake from cmake3 by default. Therefore, we need to create the link as well.
+# devtoolset-8-gcc-c++ is required to compile azure sdk in pg azure storage project
+RUN [[ %%os%% != centos ]] || [[ %%release%% != 7 ]] || ( \
+    yum -y install  perl-IPC-Cmd libuuid-devel cmake3 && \
+    ln -s /usr/bin/cmake3 /usr/bin/cmake  && \
+    yum -y remove git && \
+    yum -y install https://packages.endpointdev.com/rhel/7/os/x86_64/endpoint-repo.x86_64.rpm && \
```
hanefi, line 107: I do not like how you installed a packaging repo here.

1. This does not follow official instructions from the owners.
2. There is not enough explanation on what endpoint repo is, and why it is necessary.
3. You do not verify the signing key, and this creates a vulnerability.
4. I'd be happier to see this repository removed after we install the minimal set of packages that are absent in official CentOS repositories.

templates/Dockerfile-rpm.tmpl
```diff
@@ -93,6 +93,20 @@ RUN { \
...
+    yum -y install git && \
+    yum -y install  devtoolset-8-gcc-c++ && scl enable devtoolset-8 bash )
```
hanefi, line 109: Earlier in the same file, we installed `devtoolset-8-gcc` package only on distros that has access to that. Now you install a package that depends on this one without checking if it is actually available.

hanefi, line 109: missing `yum clean all` at the end of this command.

templates/Dockerfile-rpm.tmpl
hanefi, line 97: nit: fix extra spaces.

```diff
-RUN yum -y install  perl-IPC-Cmd libuuid-devel cmake3
+RUN yum -y install perl-IPC-Cmd libuuid-devel cmake3
```

hanefi (commented):
I see that this PR is already merged. I still wanted to share some reviews that would help increase the quality here, in the hopes that they get fixed soon.

gurkanindibay:
@hanefi  Opened an issue about the comments 
https://github.com/citusdata/packaging/issues/966

PR #975, opened 2022-08-23
Bump redhat-hll to 2.17

gurkanindibay:


postgres-matrix.yml
```diff
@@ -3,4 +3,4 @@ project_name: hll
 
 version_matrix:
   - 2.16:
```
hanefi, line 5: Version 2.16 does not support PG15. 2.17 does.

hanefi: Also we need PG10 here as we wish to create packages for that PG version as well. Do we have the necessary docker images for PG10?

hanefi (changes requested):

postgres-matrix.yml
gurkanindibay: Fixed wrong mapping and added pg 10

onurctirtir (approved):

PR #977, opened 2022-08-25
Add support for building repos with git submodules

gledis69:
This is my attempt at adding support to build repos which rely on git submodules.
The change is primarily a short script which gets executed only if the repo being built contains a .gitmodules file.
This should not affect current citus related packaging logic at all.

@hanefi @onurctirtir  since there is no released version yet, build package fails. Please ignore these errors

### DIFF
--- a/scripts/fetch_and_build_rpm
+++ b/scripts/fetch_and_build_rpm
@@ -168,7 +168,7 @@ tar czf "${tarballpath}" "${pkgsrcdir}"
 
 # git metadata needs to be setup to initialize submodules
 # in repos which rely on git submodules
-if [[ $(ls -a "${pkgsrcdir}" | grep .gitmodules | wc -l) -gt 0 ]]; then
+if [[ -z "${pkgsrcdir}/.gitmodules" ]]; then
     setup_submodules "${pkgsrcdir}"
 fi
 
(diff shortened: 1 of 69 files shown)


scripts/fetch_and_build_rpm
```diff
@@ -166,6 +166,12 @@ tarballpath="${rpmbuilddir}/${gitsha}"
 tar xf "${download}" -C "${pkgsrcdir}" --strip-components 1
 tar czf "${tarballpath}" "${pkgsrcdir}"
 
+# git metadata needs to be setup to initialize submodules
+# in repos which rely on git submodules
+if [[ $(ls -a "${pkgsrcdir}" | grep .gitmodules | wc -l) -gt 0 ]]; then
```
gledis69: This is a shorter way to check for the file I guess
```suggestion
if [[ -z "${pkgsrcdir}/.gitmodules" ]]; then
```

scripts/setup_submodules
```diff
@@ -0,0 +1,33 @@
...
+# slightly modified from: https://stackoverflow.com/a/53899440/15053670
+
+set -euxo pipefail
+
+REPO_PATH=${1:-.}
+cd $REPO_PATH
+
+git init
+
+git config -f .gitmodules --get-regexp '^submodule\..*\.path$' |
```
gurkanindibay, line 15: @gledis69 Could you add some more comment it is hard to grasp.

gledis69: Comments added, git doesn't detect them since only line 15 was selected for this comment.
Can you check if those help some more.

gurkanindibay (approved):

PR #981, opened 2022-09-16
Adds pg15 into support matrix for 11.1.1

gurkanindibay:


.github/workflows/build-citus-community-nightlies.yml
```diff
@@ -25,7 +25,6 @@ jobs:
         platform:
           - el/7
           - el/8
-          - ol/7
```
onurctirtir, line 28: why we're removing those ?

gurkanindibay: ol/7 packages does not exist in PGDG for now so we could not bake an image

onurctirtir (approved):

PR #984, opened 2022-09-21
Adds pgazure packaging image changes for deb packages

gurkanindibay:
Pg azure project is using build technologies different from other citus applications. 
Differences
1. Uses git submodule to be able to fetch submodules from their git repos.
2. Uses cmake and vpkg to be able to fetch and build from source of their dependencies

We already added integration for rpm packages. With this PR, we aim to add support for deb packages

### DIFF
--- a/templates/Dockerfile-deb.tmpl
+++ b/templates/Dockerfile-deb.tmpl
@@ -71,9 +71,7 @@ RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ %%release%%-pgdg main 15'
         liblz4-dev \
         liblz4-1 \
         libzstd1 \
-        libzstd-dev \
-        sudo \
-    && rm -rf /var/lib/apt/lists/*
+        libzstd-dev
 
 RUN pip3 install sphinx
 
@@ -83,12 +81,33 @@ RUN curl -sL https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64 
     && chmod +x /usr/bin/jq
 
 # install packagecloud repos for pg_auto_failover
-RUN curl -s https://packagecloud.io/install/repositories/citusdata/community/script.deb.sh | bash \
-    && rm -rf /var/lib/apt/lists/*
+RUN curl -s https://packagecloud.io/install/repositories/citusdata/community/script.deb.sh | bash
 
 # patch pg_buildext to use multiple processors
 COPY make_pg_buildext_parallel.patch /
 RUN patch `which pg_buildext` < /make_pg_buildext_parallel.patch
+
+# install cmake dependencies and pgazure build dependencies
+RUN apt install -y  checkinstall \
+                    git \
+                    libtemplate-perl \
+                    make \
+                    pkg-config \
+                    tar \
+                    unzip \
+                    uuid \
+                    uuid-dev \
+                    zip \
+                 && rm -rf /var/lib/apt/lists/*
+
+# install cmake from source
+RUN wget https://github.com/Kitware/CMake/releases/download/v3.22.2/cmake-3.22.2.tar.gz && \
+    tar -zxvf cmake-3.22.2.tar.gz && \
+    cd cmake-3.22.2 && ./bootstrap && \
+    make && \
+    make install && \
+    rm -f cmake-3.22.2.tar.gz && \
+    rm -rf cmake-3.22.2
 
 # Added for pg15 beta package support.
 ENV DEB_PG_SUPPORTED_VERSIONS="10 11 12 13 14 15"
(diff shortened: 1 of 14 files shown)


templates/Dockerfile-deb.tmpl
```diff
@@ -72,8 +79,7 @@ RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ %%release%%-pgdg main 15'
         liblz4-1 \
         libzstd1 \
         libzstd-dev \
-        sudo \
-    && rm -rf /var/lib/apt/lists/*
```
JelteF, line 76: this line should be added back again to avoid docker image bloat

templates/Dockerfile-deb.tmpl
```diff
@@ -37,6 +37,13 @@ RUN set -ex; \
     rm -rf "$GNUPGHOME"; \
     apt-key list
 
+# add buster backports repo to be able to download missing packages in buster main repo
+RUN ( [ %%os%% != debian ] || [ %%release%% != buster ] ) || ( \
+    echo "deb http://deb.debian.org/debian buster-backports main" >> /etc/apt/sources.list && \
+    apt-get update -y \
```
JelteF: This shouldn't do an update here

JelteF:
```suggestion
```

templates/Dockerfile-deb.tmpl
```diff
@@ -83,13 +89,36 @@ RUN curl -sL https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64
...
+RUN apt install -y  checkinstall \
+                    git \
+                    libtemplate-perl \
+                    make \
+                    pkg-config \
+                    tar \
+                    unzip \
+                    uuid \
+                    uuid-dev \
+                    zip \
```
JelteF: these should be made part of the other install list

templates/Dockerfile-deb.tmpl
gurkanindibay: Fixed

templates/Dockerfile-deb.tmpl
gurkanindibay: Fixed

templates/Dockerfile-deb.tmpl
gurkanindibay: Merged with upper block

templates/Dockerfile-deb.tmpl
```diff
@@ -83,13 +99,24 @@ RUN curl -sL https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64
     && chmod +x /usr/bin/jq
 
 # install packagecloud repos for pg_auto_failover
-RUN curl -s https://packagecloud.io/install/repositories/citusdata/community/script.deb.sh | bash \
-    && rm -rf /var/lib/apt/lists/*
```
JelteF: should be added back I think

gurkanindibay: added back

templates/Dockerfile-deb.tmpl
```diff
@@ -72,6 +77,17 @@ RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ %%release%%-pgdg main 15'
         liblz4-1 \
         libzstd1 \
         libzstd-dev \
+# below are needed for cmake and pgazure build
+        checkinstall \
+        git \
+        libtemplate-perl \
+        make \
```
hanefi: It is really strange to see an explicit installation step for a basic tool as `make` that we usually take for granted.

We already have `build-essential` listed above, and AFAIK it always depends on `make` so we can skip this one. However there is no harm in adding it to the list.

templates/Dockerfile-deb.tmpl
```diff
@@ -90,6 +106,18 @@ RUN curl -s https://packagecloud.io/install/repositories/citusdata/community/scr
 COPY make_pg_buildext_parallel.patch /
 RUN patch `which pg_buildext` < /make_pg_buildext_parallel.patch
 
+
+# install cmake from source
```
hanefi, line 109: Why do we need to install this from source? Can you add a comment on why it is not sufficient to just install a package?

In case CMake releases a patch version, we will not get that now.

gurkanindibay: For vpkg to work minimum citus 3.22 is required

templates/Dockerfile-deb.tmpl
```diff
@@ -37,6 +37,11 @@ RUN set -ex; \
     rm -rf "$GNUPGHOME"; \
     apt-key list
 
+# add buster backports repo to be able to download missing packages in buster main repo
+RUN ( [ %%os%% != debian ] || [ %%release%% != buster ] ) || ( \
+    echo "deb http://deb.debian.org/debian buster-backports main" >> /etc/apt/sources.list  \
+    )
```
hanefi, line 43: Note: Debian buster EOL'd on August 22'. It is bound to get more problematic to install packages in the future for this distro.

However I am ok to keep this as is, as the image builds fine for now apparently.

hanefi (approved):

PR #986, opened 2022-09-28
Enables deb support for pg_azure_storage nightlies

gurkanindibay:
Additionally tools version is upgraded into latest


hanefi (approved):
PR looks ok. I have several ideas that would make PRs like this one cleaner in the future:

- #948 so that we do not need to update versions in 6 places instead of 1.
- Go through GitHub Action workflow names and have a consistent naming scheme.

PR #997, opened 2022-10-25
Rollbacks postgres 15 testing changes

gurkanindibay:


hanefi (commented):
How about adding back ol/7 support?

The comments indicate that we uncommented them because of missing pg15 beta packages:

> temporarily removed from nightlies since pg15 beta is not being downloaded for ol/7

gurkanindibay:
I will handle this issue with the issue below. Both have some installation and build problems
https://github.com/citusdata/packaging/issues/999

hanefi (approved):
Please update comments with a brief explanation on why ol7 is commented out.

PR #998, opened 2022-10-26
Updates tools versions into v0.8.21

gurkanindibay:


hanefi (approved):
I dislike having to spend time to check if we updated all necessary tools versions. Please implement https://github.com/citusdata/packaging/issues/948

PR #1001, opened 2022-10-28
Updates  postgres version in repo installation scripts to 15

gurkanindibay:
Tested it for ol/7 and ubuntu/focal it worked

### DIFF
--- a/community-nightlies/rpm.sh
+++ b/community-nightlies/rpm.sh
@@ -36,7 +36,7 @@ pgdg_check ()
   if yum list -q postgresql15-server &> /dev/null; then
     echo "Detected postgresql15-server..."
   else
-    echo -n "Installing pgdg14 repo... "
+    echo -n "Installing pgdg15 repo... "
 
     if [ "${dist}" = "8" ]; then
       dnf -qy module disable postgresql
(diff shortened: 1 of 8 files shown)


community-nightlies/rpm.sh
```diff
@@ -32,20 +32,20 @@ curl_check ()
...
 {
-  echo "Checking for postgresql14-server..."
-  if yum list -q postgresql14-server &> /dev/null; then
-    echo "Detected postgresql14-server..."
+  echo "Checking for postgresql15-server..."
+  if yum list -q postgresql15-server &> /dev/null; then
+    echo "Detected postgresql15-server..."
   else
-    echo -n "Installing pgdg14 repo... "
+    echo -n "Installing pgdg15 repo... "
```
hanefi: I think it makes more sense to drop 15 from here. Pgdg no longer has separate repos for each pg version.

Please fix all the other occurences in other files as well

hanefi (dismissed):
One comment that would be really nice to have. Please consider that before merging

community-nightlies/rpm.sh
gurkanindibay: Fixed as you suggested

PR #1006, opened 2022-11-11
Adds almalinux 9 as el/ol 9 build distro

gurkanindibay:
Previously, we were using centos docker images as our container image to build el and ol packages
However, after centos 9 stream we can not build our packages since it has a whole new structure, upstream, which prevents us to use as base build image
I used almalinux as build image. You can get details from the issue below 
https://github.com/citusdata/citus/issues/6467

### DIFF
--- a/update_dockerfiles
+++ b/update_dockerfiles
@@ -35,6 +35,7 @@ function update_extra_repositories_and_packages {
             extra_repositories='oracle-epel-release-el8'
             extra_packages='llvm-toolset ccache spectool curl'
             ;;
+        # almalinux already has curl installed and installing it again causes conflicting installation error
         almalinux+9)
             extra_repositories='epel-release'
             extra_packages='llvm-toolset ccache rpmdevtools krb5-devel'
--- a/update_dockerfiles
+++ b/update_dockerfiles
@@ -21,19 +21,23 @@ function update_extra_repositories_and_packages {
     case ${os}+${release} in
         centos+6|centos+7)
             extra_repositories='epel-release centos-release-scl-rh'
-            extra_packages='llvm-toolset-7-clang llvm5.0'
+            extra_packages='llvm-toolset-7-clang llvm5.0 spectool curl'
             ;;
         centos+8)
             extra_repositories='epel-release'
-            extra_packages='llvm-toolset ccache'
+            extra_packages='llvm-toolset ccache spectool curl'
             ;;
         oraclelinux+6|oraclelinux+7)
             extra_repositories=''
-            extra_packages='llvm-toolset-7-clang llvm5.0'
+            extra_packages='llvm-toolset-7-clang llvm5.0 spectool curl'
             ;;
         oraclelinux+8)
             extra_repositories='oracle-epel-release-el8'
-            extra_packages='llvm-toolset ccache'
+            extra_packages='llvm-toolset ccache spectool curl'
+            ;;
+        almalinux+9)
+            extra_repositories='epel-release'
+            extra_packages='llvm-toolset ccache rpmdevtools krb5-devel'
             ;;
     esac
 }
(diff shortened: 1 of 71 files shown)


update_dockerfiles
```diff
@@ -21,19 +21,23 @@ function update_extra_repositories_and_packages {
...
             ;;
         centos+8)
             extra_repositories='epel-release'
-            extra_packages='llvm-toolset ccache'
+            extra_packages='llvm-toolset ccache spectool curl'
             ;;
         oraclelinux+6|oraclelinux+7)
             extra_repositories=''
-            extra_packages='llvm-toolset-7-clang llvm5.0'
+            extra_packages='llvm-toolset-7-clang llvm5.0 spectool curl'
```
onurctirtir, line 32: we did we need those two at the end ?

onurctirtir (commented):
LGTM except one comment.

update_dockerfiles
gurkanindibay: We should evaluate these changes with https://github.com/citusdata/packaging/blob/c4ddbb4105cc8dab6d8aaf3467a1272c9f244af7/templates/Dockerfile-rpm.tmpl#L65
For Almalinux, these packages are unnecessary and cause an error since repos do not include these packages.
therefore, I made these packages as extra packages and did not add these packages as extra packages for Almalinux

onurctirtir: I see your point now.

So my next question is; if these packages are unnecessary for some distros, could that be the case that they are unnecessary for all ? I mean, what happens if we not install them for any of those distros ?

gurkanindibay: We will get error when baking docker images. Therefore, we will get error in pipeline

onurctirtir: Could you share some more details about the error (maybe copy-paste it ?), I would like to understand the situation better.

gurkanindibay: While executing the command with packages spectool and curl in Dockerfile
```
docker build -t citusdata/packaging-test:almalinux-9-15 -f dockerfiles/almalinux-9-pg15/Dockerfile .
```
Complete!
Last metadata expiration check: 0:01:56 ago on Thu Nov 17 07:38:27 2022.
Package flex-2.6.4-9.el9.x86_64 is already installed.
Package gcc-c++-11.3.1-2.1.el9.alma.x86_64 is already installed.
Package libstdc++-devel-11.3.1-2.1.el9.alma.x86_64 is already installed.
Package rpm-build-4.16.1.3-17.el9.x86_64 is already installed.
Package tar-2:1.34-3.el9.x86_64 is already installed.
Package libzstd-1.5.1-2.el9.x86_64 is already installed.
Package libzstd-devel-1.5.1-2.el9.x86_64 is already installed.
**No match for argument: spectool
Error: Unable to find a match: spectool**

gurkanindibay: Error:
 Problem: problem with installed package curl-minimal-7.76.1-14.el9_0.5.x86_64
  - package curl-minimal-7.76.1-14.el9_0.5.x86_64 conflicts with curl provided by curl-7.76.1-19.el9.x86_64
  - package curl-minimal-7.76.1-19.el9.x86_64 conflicts with curl provided by curl-7.76.1-19.el9.x86_64
  - conflicting requests

update_dockerfiles
```diff
@@ -21,19 +21,23 @@ function update_extra_repositories_and_packages {
     case ${os}+${release} in
```
onurctirtir, line 21: Maybe add such a comment somewhere over here:
`almalinux has "curl" already so we need to install it only for other distros`

onurctirtir (approved):
I'm still wondering do we really need to have `spectools` for other distros, but this can be fixed later

PR #1009, opened 2022-11-18
Adds ubuntu-kinetic support

gurkanindibay:


hanefi (changes requested):
The scripts in `scripts` directory have missing `+x` permissions. I think they should be 755 instead of 644.

The same mistake is there on many other files. The template files for these scripts have the correct permissions but somehow they are wrong on many of the other directories such as `ubuntu-kinetic-all/scripts`.

I suggest that you do it correctly, and create a mechanism to check that they are not broken again.

gurkanindibay:
> The scripts in `scripts` directory have missing `+x` permissions. I think they should be 755 instead of 644.
> 
> The same mistake is there on many other files. The template files for these scripts have the correct permissions but somehow they are wrong on many of the other directories such as `ubuntu-kinetic-all/scripts`.
> 
> I suggest that you do it correctly, and create a mechanism to check that they are not broken again.

I removed all the scripts under dockerfile directory and updated update_dockerfile not to update the useless scripts anymore,
since scripts under ./scripts directory are actually being copied into docker images in update_images script

hanefi:
> scripts under ./scripts directory are actually being copied into docker images in update_images script

Can you share a link to the lines of code that does this? I didn't know we did that.

https://github.com/citusdata/packaging/blob/2fbc111819be6a28367885d2a1ed8ae4bd0d9d97/update_image#L50

You can see that docker build is being executed under the top level directory. Therefore, script files under ./scripts directory actually being copied into images

hanefi (approved):

PR #1020, opened 2022-12-26
Adds python3.8 installations in Docker files

gurkanindibay:
Python3.8 is minimum requirement for tools scripts. However, ubuntu bionic, debian buster, centos and oraclelinux releases have python releases lower than 3.8. 
In citus pipelines we are executing tools scripts inside image scripts. Therefore, python 3.8 installation is a must for above release pairs

### DIFF
--- a/templates/Dockerfile-deb.tmpl
+++ b/templates/Dockerfile-deb.tmpl
@@ -124,6 +124,9 @@ RUN !( ( [ %%os%% = debian ] && [ %%release%% = buster ] ) || \
         tar -xvf Python-3.8.16.tar.xz && \
         cd Python-3.8.16 && \
         ./configure --enable-optimizations && \
+        make -j 8 && \
+        make altinstall && \
+        cd .. && \
         rm -f Python-3.8.16.tar.xz && \
         rm -rf Python-3.8.16  \
      )
--- a/templates/Dockerfile-deb.tmpl
+++ b/templates/Dockerfile-deb.tmpl
@@ -119,7 +119,7 @@ RUN wget https://github.com/Kitware/CMake/releases/download/v3.22.2/cmake-3.22.2
 ARG PYTHON_VERSION=3.8.16
 RUN set -ex \
     && curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash \
-    && export PYENV_ROOT="$HOME/.pyenv" && export PATH="$PYENV_ROOT/bin:$PATH" \
+    && export PATH="$HOME/.pyenv/bin:$PATH" \
     && pyenv update \
     && pyenv install $PYTHON_VERSION \
     && pyenv global $PYTHON_VERSION \
--- a/templates/Dockerfile-deb.tmpl
+++ b/templates/Dockerfile-deb.tmpl
@@ -125,7 +125,10 @@ RUN set -ex \
     && pyenv update \
     && pyenv install $PYTHON_VERSION \
     && pyenv global $PYTHON_VERSION \
-    && pyenv rehash
+    && pyenv rehash \
+    && echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bashrc \
+    && echo 'command -v pyenv >/dev/null || export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bashrc \
+    && echo 'eval "$(pyenv init -)"' >> ~/.bashrc
 
 
 # Added for pg15 beta package support.
--- a/templates/Dockerfile-deb.tmpl
+++ b/templates/Dockerfile-deb.tmpl
@@ -120,12 +120,6 @@ RUN wget https://github.com/Kitware/CMake/releases/download/v3.22.2/cmake-3.22.2
 RUN !( ( [ %%os%% = debian ] && [ %%release%% = buster ] ) || \
         ( [ %%os%% = ubuntu ] && [ %%release%% = bionic ] ) \
      ) || ( \
-        apt-get install -y  libncurses5-dev  \
-                            libgdbm-dev \
-                            libnss3-dev \
-                            libsqlite3-dev \
-                            libffi-dev \
-                            libbz2-dev \
         wget https://www.python.org/ftp/python/3.8.16/Python-3.8.16.tar.xz && \
         tar -xvf Python-3.8.16.tar.xz && \
         cd Python-3.8.16 && \
--- a/templates/Dockerfile-deb.tmpl
+++ b/templates/Dockerfile-deb.tmpl
@@ -120,6 +120,12 @@ RUN wget https://github.com/Kitware/CMake/releases/download/v3.22.2/cmake-3.22.2
 RUN !( ( [ %%os%% = debian ] && [ %%release%% = buster ] ) || \
         ( [ %%os%% = ubuntu ] && [ %%release%% = bionic ] ) \
      ) || ( \
+        apt-get install -y  libncurses5-dev  \
+                            libgdbm-dev \
+                            libnss3-dev \
+                            libsqlite3-dev \
+                            libffi-dev \
+                            libbz2-dev \
         wget https://www.python.org/ftp/python/3.8.16/Python-3.8.16.tar.xz && \
         tar -xvf Python-3.8.16.tar.xz && \
         cd Python-3.8.16 && \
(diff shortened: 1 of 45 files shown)


templates/Dockerfile-deb.tmpl
```diff
@@ -115,6 +115,23 @@ RUN wget https://github.com/Kitware/CMake/releases/download/v3.22.2/cmake-3.22.2
     rm -f cmake-3.22.2.tar.gz && \
     rm -rf cmake-3.22.2
 
+# install python 3.8 to be able to execute tools scripts
+# if os/release in (debian/buster,ubuntu/bionic)
+RUN !( ( [ %%os%% = debian ] && [ %%release%% = buster ] ) || \
+        ( [ %%os%% = ubuntu ] && [ %%release%% = bionic ] ) \
+     ) || ( \
```
hanefi:
```suggestion
     ) && ( \
```

It looks like we install python 3.8 for all the distros other than buster and bionic.The comments indicate that we should do it other way around.

templates/Dockerfile-rpm.tmpl
```diff
@@ -105,6 +105,28 @@ RUN [[ %%os%% != centos ]] || [[ %%release%% != 7 ]] || ( \
     yum -y install git && \
     yum -y install  devtoolset-8-gcc-c++ && scl enable devtoolset-8 bash )
 
+# install python 3.8 to be able to execute tools scripts
+# if os in (centos, oraclelinux) and release in (7, 8)
+RUN !( ( [ %%os%% = centos ] && [ %%release%% = 7 ] ) || \
+       ( [ %%os%% = centos ] && [ %%release%% = 8 ] ) || \
+       ( [ %%os%% = oraclelinux ] && [ %%release%% = 7 ] ) || \
+       ( [ %%os%% = oraclelinux ] && [ %%release%% = 8 ] ) \
```
hanefi:
```suggestion
RUN !( ( [ %%os%% = centos ] && [ %%release%% <= 8 ] ) || \
       ( [ %%os%% = oraclelinux ] && [ %%release%% <= 8 ] ) \
```

this is easier to read IMO. this is optional and you can ignore this comment.

templates/Dockerfile-rpm.tmpl
```diff
@@ -105,6 +105,28 @@ RUN [[ %%os%% != centos ]] || [[ %%release%% != 7 ]] || ( \
...
+     ) || ( \
```
hanefi:
```suggestion
     ) && ( \
```

I think this is a typo here.

hanefi (changes requested):

hanefi:
An alternative solution that would be cleaner than this: https://github.com/pyenv/pyenv

You can install and set the desired python version using a single command.

templates/Dockerfile-rpm.tmpl
gurkanindibay: Installed pyenv for all distros. Therefore no need to change this. BTW I tested this condition and it works as intended

templates/Dockerfile-deb.tmpl
gurkanindibay: Installed pyenv for all distros. Therefore no need to change this. BTW I tested this condition and it works as intended

hanefi: Looks like I missed the negation in `RUN !(`. Thanks for not implementing my suggestion blindly.

templates/Dockerfile-deb.tmpl
```diff
@@ -115,6 +116,17 @@ RUN wget https://github.com/Kitware/CMake/releases/download/v3.22.2/cmake-3.22.2
...
+RUN git clone https://github.com/pyenv/pyenv.git ~/.pyenv && \
+    echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bashrc && \
+    echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bashrc && \
+    echo -e 'if command -v pyenv 1>/dev/null 2>&1; then\n eval "$(pyenv init -)"\nfi' >> ~/.bashrc
+
+# install python 3.8 and set it as default. Reloading the bashrc did not worked for ubuntu \
+# therefore I need to export the path explicitly here.
+RUN source ~/.bashrc && export PYENV_ROOT="$HOME/.pyenv" && export PATH="$PYENV_ROOT/bin:$PATH" && \
+    pyenv install 3.8.16 && pyenv global 3.8.16
+
```
hanefi:
```suggestion
# Install pyenv
ARG PYTHON_VERSION=3.8.16
RUN set -ex \
    && curl https://pyenv.run | bash \
    && pyenv update \
    && pyenv install $PYTHON_VERSION \
    && pyenv global $PYTHON_VERSION \
    && pyenv rehash
```

This online shell script (https://pyenv.run) actually sets the necessary variables, downloads the git repo**s** when run.

You can apply a similar patch for rpm as well. It will make it unnecessary to use bash.

hanefi: You can also move the `ARG` line to earlier in the file.

gurkanindibay: @hanefi I think it is not good to rely on some domain name we don't know when we look at in security perspective. 
When I make a whois search, I see that this domain name owner is RadiaSoft LLC which is an unknown company.
However, parameterization of Python version is good idea. I need to handle that
What do you think?

hanefi: How about the following?

`curl -s -S -L https://raw.githubusercontent.com/pyenv/pyenv-installer/master/bin/pyenv-installer | bash`

That link that is hosted by RadiaSoft is actually a one-liner that runs the most up-to-date pyenv-installer script. Do you think that it is safe to install a file from a well known Github repo and execute it?

hanefi: See the following docs from pyenv-installer: https://github.com/pyenv/pyenv-installer#using-pyenv-installer-on-travis-ci

templates/Dockerfile-rpm.tmpl
```diff
@@ -105,6 +104,19 @@ RUN [[ %%os%% != centos ]] || [[ %%release%% != 7 ]] || ( \
     yum -y install git && \
     yum -y install  devtoolset-8-gcc-c++ && scl enable devtoolset-8 bash )
 
+# install python 3.8 to be able to execute tools scripts
+ARG PYTHON_VERSION=3.8.16
+RUN yum -y install  bzip2-devel \
+                libffi-devel \
+                xz-devel && \
```
hanefi: An earlier `yum clean all` command cleaned caches and now this command requires to download repo information one more time.

I can see 2 options:
1.  Moving installation of these new packages to earlier in the file, merging them with other package installations.
2. Adding a `yum clean all` somewhere in this `RUN` command to decrease docker image and cache sizes.

templates/Dockerfile-deb.tmpl
gurkanindibay: Changed as you suggested

templates/Dockerfile-rpm.tmpl
gurkanindibay: Moved package installation into upper package installation blocks

hanefi (approved):

templates/Dockerfile-deb.tmpl
```diff
@@ -115,6 +117,18 @@ RUN wget https://github.com/Kitware/CMake/releases/download/v3.22.2/cmake-3.22.2
...
+RUN set -ex \
+    && curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash \
+    && export PATH="$HOME/.pyenv/bin:$PATH" \
+    && pyenv update \
+    && pyenv install $PYTHON_VERSION \
+    && pyenv global $PYTHON_VERSION \
+    && pyenv rehash \
+    && echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bashrc \
+    && echo 'command -v pyenv >/dev/null || export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bashrc \
+    && echo 'eval "$(pyenv init -)"' >> ~/.bashrc
```
gurkanindibay, line 131: Added commands to be able to call pyenv directly

templates/Dockerfile-deb.tmpl
```diff
@@ -55,6 +55,8 @@ RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ %%release%%-pgdg main' >
         devscripts \
         fakeroot \
         flex \
+        libbz2-dev \
+        libffi-dev \
```
gurkanindibay, line 59: Added packages to extract zip files and source build pyenv

hanefi (approved):
I do not like the fact that we assume the shell is bash. However after running some tests I saw that the interpreters for entrypoint scripts are all bash, so it looks like we are ok here.

PR #1021, opened 2023-01-04
Adds ubuntu jammy support for 10.2.8

gurkanindibay:


.github/workflows/build-citus-enterprise-nightlies.yml
```diff
@@ -29,6 +29,7 @@ jobs:
           - debian/bullseye
           - ubuntu/bionic
           - ubuntu/focal
+          - ubuntu/jammy
```
JelteF: I think we can simply remove this action. We're shouldn't be building enterprise nigthlies anymore now that we're not changing master branch of enterprise repo, because everything is open source.

gurkanindibay: Removed

hanefi (approved):

PR #1036, opened 2023-04-04
Bump debian pg_cron to 1.5.1

gurkanindibay:


debian/changelog
```diff
@@ -1,3 +1,9 @@
+pgcron (1.5.1.citus-1) stable; urgency=low
+
+  * Official 1.5.1 release of Pg Cron
```
hanefi: This is a weird way to write an abbreviation for PostgreSQL. How about dropping the name altogether, or capitalizing PG part?

```suggestion
  * Official 1.5.1 release
```
or
```suggestion
  * Official 1.5.1 release of PG Cron
```

debian/changelog
```diff
@@ -1,3 +1,9 @@
...
+
+ -- Gurkan Indibay <marco@citusdata.com>  Thu, 4 Apr 2022 10:55:45 +0000
```
hanefi: The name is yours, but email belongs to Marco. This feels weird.

hanefi (changes requested):

debian/changelog
gurkanindibay: Fixed with the second alternative

debian/changelog
gurkanindibay: Sorry for the confusion. Fixed email address

hanefi (approved):

PR #1041, opened 2023-04-10
Adds changes to packaging images to package pg_azure 1.2.0

gurkanindibay:
There are two category of changes in this PR:
1. Changes required for pg_azure build
2. Changes because of entropy in the packaging system

Two changes are incorporated in docker images to build pg_azure 1.2.0:

1. Three new environment variables are secret and should be defined in GH Actions
These environment variables are added DEB_BUILD_OPTIONS to be able to used
with packaging environment as environment variables
2. clang is required for Rust installation to work properly. clang 5.0 is added with 
the installation of llvm-toolset-7

Three changes are incorporated to fix the entropy in packaging images:

1. perl-IPC-Run is the new dependency for postgresql packages in
rpm based systems. perl-IPC-Run is installed for all rpm based distros.
Since this package is not a package resides in standard repositories, it is 
needed to be installed in different ways for different os/release pairs
2. sphinx was being installed using pip. However, it was failing for debian
bookworm because of a permission issue. Direct pip installation is replaced
with python3-sphinx for which Docker environment has permission
4. Postgres 10 is removed from GH Actions matrix for rpm since Postgres 10 packages 
does not exist any more in pgdg repositories


.github/workflows/build-package.yml
```diff
@@ -43,7 +43,6 @@ jobs:
           - oraclelinux,7
           - almalinux,9
         POSTGRES_VERSION:
-          - 10
```
gurkanindibay, line 46: Postgres 10 packages are not being served any more for rpm based distros from pgdg repository. Therefore I removed postgres 10
For reference, https://github.com/citusdata/packaging/actions/runs/4659305909

.github/workflows/build-pgazure-nightlies.yml
```diff
@@ -10,9 +10,7 @@ env:
   DOCKERHUB_USER_NAME: ${{ secrets.DOCKERHUB_USER_NAME }}
   DOCKERHUB_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
 on:
-  push:
```
gurkanindibay, line 13: pg_azure failing right now, since packaging images are not ready and since this pipeline is being executing on each commit, it was blocking PR to fix pg_azure packaging. Therefore, there was a circular dependency and I removed it

templates/Dockerfile-rpm.tmpl
```diff
@@ -7,6 +7,31 @@ RUN [[ %%os%% != centos ]] || [[ %%release%% != 8 ]] || ( \
 
 RUN yum -y update
 
+RUN  [[ %%os%% != centos ]] || [[ %%release%% != 8 ]] || ( \
```
gurkanindibay, line 10: perl-IPC-Run is a new dependency for rpm based distros. Installation is not straightforward. I needed to perform different steps for different os/release pair
For oraclelinux, there is no repo including this package. Therefore I needed to install manually

templates/Dockerfile-rpm.tmpl
```diff
@@ -105,7 +130,10 @@ RUN [[ %%os%% != centos ]] || [[ %%release%% != 7 ]] || ( \
     yum -y remove git && \
     yum -y install https://packages.endpointdev.com/rhel/7/os/x86_64/endpoint-repo.x86_64.rpm && \
     yum -y install git && \
-    yum -y install  devtoolset-8-gcc-c++ && scl enable devtoolset-8 bash )
+    yum -y install  devtoolset-8-gcc-c++ && scl enable devtoolset-8 bash && \
+    yum -y install llvm-toolset-7 && \
```
gurkanindibay, line 134: Ruby required clang > 5.0 and standard clang package in repositories includes 3.x. Therefore, I installed llvm-toolset-7

templates/Dockerfile-deb.tmpl
```diff
@@ -74,6 +74,7 @@ RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ %%release%%-pgdg main' >
         wget \
         zlib1g-dev \
         python3-pip \
+        python3-sphinx \
```
gurkanindibay, line 77: pip3 install sphinx is getting error on debian bookworm and complains about insufficient permission to change system python. Suggested solution was to install pythpn3-sphinx for which docker is permitted to perform

scripts/fetch_and_build_deb
```diff
@@ -209,7 +209,15 @@ pg_buildext updatecontrol
 procs="$(nproc)"
 mjobs="$(expr $procs + 1)"
 
-DEB_BUILD_OPTIONS="parallel=${mjobs}" debuild --prepend-path /usr/local/bin --preserve-envvar CONF_EXTRA_VERSION --preserve-envvar UNENCRYPTED_PACKAGE --preserve-envvar PACKAGE_ENCRYPTION_KEY -uc -us -B --lintian-opts --profile debian --allow-root
+DEB_BUILD_OPTIONS="parallel=${mjobs}" debuild \
+                  --prepend-path /usr/local/bin \
```
gurkanindibay, line 213: Added three new environment for debian packages to be preserved in packaging environment. Other 2 environment variables, INSTALL_RUST and CI will be set in pg_azure packaging configuration files, since they are fixed and not secret. Therefore, we don't need to set them on Github Actions.

.github/workflows/build-package-test.yml
```diff
@@ -12,8 +12,8 @@ env:
   CURRENT_BRANCH: ${GITHUB_REF##*/}
   TEST: true
 on:
-  push:
-    branches: "**"
+#  push:
+#    branches: "**"
```
hanefi: We should uncomment these lines before merging the PR.

```suggestion
  push:
    branches: "**"
```

.github/workflows/build-package.yml
hanefi: I am ok with removing PG10. However, removing from matrix is not enough. We should remove all files and directories that are for PG10 as well.

hanefi (changes requested):

.github/workflows/build-package-test.yml
gurkanindibay: Uncommented

.github/workflows/build-package.yml
gurkanindibay: Removed all the files and changed scripts as well

hanefi (approved):

PR #1043, opened 2023-04-13
Adds fallback address for debian releases 

gurkanindibay:
Recently, we have problems to download debian/bookworm 
packages from packagecloud, since packagecloud repository 
config file does not work with release number but works 
with release code name. 
The one that is not working
https://packagecloud.io/install/repositories/citusdata/community/config_file.list?os=debian&dist=12&source=script
The one that is working
https://packagecloud.io/install/repositories/citusdata/community/config_file.list?os=debian&dist=bookworm&source=script

There are two fixes in this PR to handle this problem
1. Adds new debian releases in the codename matrix for debian
2. Adds fallback address for debian releases which 
includes code names instead of release numbers.

### DIFF
--- a/community/deb.sh
+++ b/community/deb.sh
@@ -258,9 +258,15 @@ main ()
 
   echo -n "Installing $apt_source_path... "
 
-  # create an apt config file for this repository
   curl -sSf "${apt_config_url}" > $apt_source_path
   curl_exit_code=$?
+  if [ "${os}" = "debian" ] && [ "$curl_exit_code" -ne "0" ]; then
+    apt_config_url_with_code_name=("https://repos.citusdata.com/${repo_name}/config_file.list?os=${os}&dist=${codename}&source=script"
+    curl -sSf "${apt_config_url_with_code_name}" > "$apt_source_path"
+    echo "Using fallback url: ${apt_config_url_with_code_name}"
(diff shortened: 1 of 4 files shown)


community/deb.sh
```diff
@@ -249,9 +258,15 @@ main ()
 
   echo -n "Installing $apt_source_path... "
 
-  # create an apt config file for this repository
```
hanefi: Why did we remove this comment?

community/deb.sh
```diff
@@ -249,9 +258,15 @@ main ()
...
   curl -sSf "${apt_config_url}" > $apt_source_path
   curl_exit_code=$?
+  if [ "${os}" = "debian" ] && [ "$curl_exit_code" -ne "0" ]; then
```
hanefi, line 277: Can we add a comment before this conditional to explain why we need such a change?

hanefi (changes requested):
After addressing the comments, can you apply the same patch to other 3 `deb.sh` files as well?

hanefi (dismissed):

community/deb.sh
gurkanindibay: Removed mistakenly. Added again

hanefi (approved):

PR #1048, opened 2023-04-26
Adds schedule for image build health check

gurkanindibay:
Image build process may break as time passes and mostly we see problems when we start packaging which may cause delays. Therefore, to see the problem before packaging I create a schedule to build the package and see
problems beforehand
Additionally, updated tools version into latest

### DIFF
--- a/.github/workflows/build-citus-community-nightlies.yml
+++ b/.github/workflows/build-citus-community-nightlies.yml
@@ -48,7 +48,7 @@ jobs:
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
 
       - name: Clone tools branch
-        run: git clone -b v0.8.24 --depth=1  https://github.com/citusdata/tools.git tools
+        run: git clone -b v0.8.26 --depth=1  https://github.com/citusdata/tools.git tools
 
       - name: Clone build branch
         run: git clone -b "${MAIN_BRANCH}" --depth=1  https://github.com/citusdata/packaging.git packaging
--- a/.github/workflows/image-health-check.yml
+++ b/.github/workflows/image-health-check.yml
@@ -1,4 +1,4 @@
-name: Build Package
+name: Image Health Check
 
 env:
   GH_TOKEN: "${{ secrets.GH_TOKEN }}"
--- a/.github/workflows/image-health-check.yml
+++ b/.github/workflows/image-health-check.yml
@@ -14,6 +14,8 @@ env:
 on:
   push:
     branches: "**"
+  schedule:
+    - cron: "30 2 * * *"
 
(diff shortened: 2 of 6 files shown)


.github/workflows/build-citus-community-nightlies.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
 
       - name: Clone tools branch
-        run: git clone -b v0.8.24 --depth=1  https://github.com/citusdata/tools.git tools
```
hanefi, line 51: Why do you need to update tools version on ontouched workflows? I do not like how this looks.

If you always wish to have these versions up to date consider implementing https://github.com/citusdata/packaging/issues/948

.github/workflows/image-health-check.yml
```diff
@@ -0,0 +1,90 @@
...
+      fail-fast: false
+      matrix:
+        TARGET_PLATFORM:
+          - centos,8
+          - centos,7
+          - oraclelinux,8
+          - oraclelinux,7
+          - almalinux,9
+        POSTGRES_VERSION:
+          - 11
```
hanefi, line 48: Does this mean that we will try to create Citus packages for PG11? It is not clear to me what we are actually testing for when we run `test_citus_package.py` later in the file.

.github/workflows/build-citus-community-nightlies.yml
gurkanindibay: I added comments in the issue #948 . I'm changing versions one by one after executing all tests for that pipeline

.github/workflows/image-health-check.yml
gurkanindibay: No. This means that we are keeping package creation images for pg 11 up to date. Still postgres serves postgres packages for pg11 so we are continuing creating docker images for pg11

.github/workflows/image-health-check.yml
```diff
@@ -0,0 +1,90 @@
+name: Image Health Check
```
hanefi, line 1: I believe it would be better to include `update_image` in the file name as well as the `name:` attribute. This workflow do not actually check the health of existing images but tests if we can build images from scratch.

hanefi (approved):
Although I am still against touching all the workflow files here I'll still approve this one to unblock you.

.github/workflows/image-health-check.yml
gurkanindibay: @hanefi I see your point however I prefer approach the naming from why. The main reason we are creating the image is checking the health. When I see the name in Github, it makes sense to me. For other people as well

PR #1061, opened 2023-05-22
Stop using repos.citusdata.com as we have ssl problems

hanefi:
Our ssl certificates expired for `repos.citusdata.com`. As a workaround, we can consider using packagecloud domains and their ssl certificates instead.


hanefi:
We fixed the TLS certificate problem for `*.citusdata.com` and this workaround is no longer necessary.

PR #1070, opened 2023-08-29
Bump debian pg_cron to 1.6.0

gurkanindibay:


postgres-matrix.yml
```diff
@@ -6,3 +6,5 @@ version_matrix:
       postgres_versions: [10, 11, 12, 13, 14]
   - 1.4.2:
       postgres_versions: [ 10, 11, 12, 13, 14, 15 ]
+  - 1.5.2:
```
hanefi: Should this be 1.6 instead?

debian/changelog
```diff
@@ -1,3 +1,10 @@
+pgcron (1.6.0.citus-1) stable; urgency=low
+
+  * Official 1.6.0 release of PG Cron
+  * Adds support for PostgreSQL 16
+
```
hanefi, line 5: We did not use to have an empty line here for earlier versions.

hanefi (changes requested):

debian/changelog
gurkanindibay: All of them are standard. Do I miss sth?

postgres-matrix.yml
gurkanindibay: Yes Fixed. thanks

debian/changelog
hanefi: It appears that my browser was rendering a strange editor view. Your changes are totally fine in this part of the file.

hanefi (approved):

PR #1072, opened 2023-09-01
Bump debian-hll to 2.18

gurkanindibay:


postgres-matrix.yml
```diff
@@ -5,4 +5,4 @@ version_matrix:
   - 2.16:
       postgres_versions: [ 10, 11, 12, 13, 14 ]
   - 2.17:
-      postgres_versions: [ 10, 11, 12, 13, 14, 15 ]
+      postgres_versions: [ 10, 11, 12, 13, 14, 15, 16 ]
```
hanefi, line 8: PG 10 is no longer supported by the community. I think it is time we remove it to prevent any possible headaches.

hanefi (changes requested):

gurkanindibay:
Removed pg 10

hanefi (approved):

PR #1073, opened 2023-09-01
Bump redhat-hll to 2.18

gurkanindibay:


hanefi (changes requested):
I think we should remove PG10 from our builds and remove references to it in pg_exclude.yml

gurkanindibay:
removed pg 10

pg_exclude.yml
```diff
@@ -1,5 +1,8 @@
 exclude:
   nightly:
     ol/7: [15]
+    el/8: [10]
```
hanefi: Why do we still need exclusion rules for PG10?

gurkanindibay: Kept it in case of previous minors but not needed as you suggested

postgres-matrix.yml
```diff
@@ -5,4 +5,6 @@ version_matrix:
   - 2.16:
       postgres_versions: [ 10, 11, 12, 13, 14 ]
   - 2.17:
-      postgres_versions: [ 10, 11, 12, 13, 14, 15 ]
+      postgres_versions: [ 10, 11, 12, 13, 14, 15, 16 ]
```
hanefi: It does not make sense to add PG16 support to 2.17. 2.18 should be the first version to support this version of PostgreSQL.

```suggestion
      postgres_versions: [ 10, 11, 12, 13, 14, 15 ]
```

hanefi (approved):

PR #1074, opened 2023-09-01
Bump debian-topn to 2.6.0

gurkanindibay:


hanefi (changes requested):
I think we should remove PG10 from our builds

gurkanindibay:
Removed pg 10

hanefi (approved):

PR #1075, opened 2023-09-01
Bump redhat-topn to 2.6.0

gurkanindibay:


postgres-matrix.yml
```diff
@@ -5,4 +5,4 @@ version_matrix:
   - 2.4.0:
       postgres_versions: [ 10, 11, 12, 13, 14 ]
   - 2.5.0:
-      postgres_versions: [ 10, 11, 12, 13, 14, 15 ]
+      postgres_versions: [ 10, 11, 12, 13, 14, 15, 16 ]
```
hanefi: PG 10 is no longer supported by the community. I think it is time we remove it to prevent any possible headaches.

pg_exclude.yml
```diff
@@ -1,5 +1,8 @@
 exclude:
   nightly:
     ol/7: [15]
+    el/8: [10]
```
hanefi: I think we should stop creating packages for PG 10

hanefi (changes requested):

pg_exclude.yml
gurkanindibay: Removed pg 10

hanefi: I still see the exclusion rules here.

hanefi:
I do not think we need exclusion rules for PG 10. Instead we should drop support for them properly

pg_exclude.yml
gurkanindibay: Kept it in case of previous minors but not needed as you suggested

postgres-matrix.yml
```diff
@@ -5,4 +5,6 @@ version_matrix:
...
```
hanefi: It does not make sense to add PG16 support to 2.5.0. 2.6 should be the first version to support this version of PostgreSQL.

```suggestion
      postgres_versions: [ 10, 11, 12, 13, 14, 15 ]
```

hanefi (approved):

PR #1078, opened 2023-09-07
Adds CRATES_IO_MIRROR_FEED_TOKEN env for pg_azure

gurkanindibay:


hanefi (approved):

scripts/fetch_and_build_deb
```diff
@@ -217,6 +217,7 @@ DEB_BUILD_OPTIONS="parallel=${mjobs}" debuild \
                   --preserve-envvar MSRUSTUP_PAT \
                   --preserve-envvar MSCODEHUB_USERNAME \
                   --preserve-envvar MSCODEHUB_PASSWORD \
```
hanefi: It might make sense to remove mscodehub env vars now.

hanefi (changes requested):

hanefi:
If it is too much work to make the tests pass, we can open an issue and move on without actually removing the preserve-env parameters.

PR #1100, opened 2024-04-24
Bump to Citus 12.1.3

gurkanindibay:
Other than regular version changes, I needed to remove sudo apt-get update commands from pipeline files since weird errors appeared releated to microsoft packages. I suppose that there may be changes in repositories of stock images of Github runners. 
 Here you can sess the error:
 https://github.com/citusdata/packaging/actions/runs/8813807672/job/24192341550#step:4:37


hanefi:
> I needed to remove sudo apt-get update commands from pipeline files

I do not think that it is acceptable to build packages with outdated package versions. There may be security patches that we may need here.

hanefi (approved):
After some offline discussion, I am convinced that there is not much risk in removing `apt-get update` commands in our workflows.

PR #1104, opened 2024-06-05
Adds el/9 ol/9 to supported os

gurkanindibay:


.github/workflows/build-package.yml
```diff
@@ -40,6 +41,9 @@ jobs:
       - name: Install python requirements
         run: python -m pip install -r tools/packaging_automation/requirements.txt
 
+      - name: Cleanup Docker
```
gurkanindibay, line 44: I was getting No disk error so that I added this step

hanefi (approved):

PR #1105, opened 2024-06-05
Adds almalinux to supported os

gurkanindibay:
Here in the tests, citus package downloaded.
Since hll and topn is not ready, I couldn't continue
https://github.com/citusdata/tools/actions/runs/9384241067/job/25839624420#step:5:132

### DIFF
--- a/community-nightlies/rpm.sh
+++ b/community-nightlies/rpm.sh
@@ -202,7 +202,7 @@ detect_repo_url ()
   pkg_dist="${dist}"
 
   case "${os}" in
-    ol|centos|rhel|redhatenterpriseserver)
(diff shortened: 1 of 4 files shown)


hanefi (commented):
Good first step. I was able to install our packages along with all the dependencies.

However, I'd like to see the same change applied to other rpm.sh files in the project.

gurkanindibay:
> Good first step. I was able to install our packages along with all the dependencies.
> 
> However, I'd like to see the same change applied to other rpm.sh files in the project.

Added to all of them

hanefi (approved):

gurkanindibay:
After adding hll and topn, we can merge it

PR #1113, opened 2024-12-23
Bump pg_cron to 1.6.5 and update build workflow for new platforms

gurkanindibay:


packaging_ignore.yml
```diff
@@ -9,6 +9,8 @@ base:
   - "gpg: WARNING: options in `/root/.gnupg/gpg.conf' are not yet active during this run"
   - ".* warning: overriding recipe for target 'check'"
   - "WARNING generated by debuild:"
+  - "Wrote: /citus-rpm-build/*"
+  - "\\d+ packages and \\d+ specfiles checked; \\d+ errors, \\d+ warnings."
```
hanefi, line 13: Shouldn't these be under rpm?

hanefi (changes requested):

packaging_ignore.yml
gurkanindibay: There is an issue related to categorizing ignore under rpm category. Therefore, I added under general category

hanefi (approved):

PR #1123, opened 2025-01-23
Bump debian-topn to 2.7.0

m3hm3t:
Bump debian-topn to 2.7.0

### DIFF
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -26,7 +26,6 @@ jobs:
           - debian/buster
           - debian/bullseye
           - debian/bookworm
-          - ubuntu/bionic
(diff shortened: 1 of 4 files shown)


.github/workflows/build-package.yml
```diff
@@ -26,7 +26,6 @@ jobs:
           - debian/buster
           - debian/bullseye
           - debian/bookworm
-          - ubuntu/bionic
```
m3hm3t, line 29: @gurkanindibay 
Should I revert this?

gurkanindibay: Change is correct no need

gurkanindibay (approved):

PR #1134, opened 2025-03-13
Bump to citus 13.0.2

ihalatci:


m3hm3t (approved):

ihalatci:
failing jobs are for nightly which runs on main and it does not have pg17 support yet.

PR #1182, opened 2026-06-16
Add PostgreSQL 16 core package build for Ubuntu focal

kemalbuyukkaya:
PGDG dropped PostgreSQL 16 binaries for Ubuntu 20.04 (focal) after focal reached EOL standard support; the last official focal build was 16.9-1.pgdg20.04+1. This adds a standalone, signed build pipeline that rebuilds newer 16.x core packages for focal.

Approach (validated locally, produces the full 13-package set that installs and runs on stock focal with working JIT):
- Combine the newer upstream orig.tar.bz2 with the focal-era debian/ packaging (16.9-1.pgdg20.04+1), whose default toolchain (clang/llvm-dev = LLVM 10) yields focal-native dependencies (libicu66, libssl1.1, libldap-2.4-2, libllvm10) rather than the clang-19/llvm-19 required by newer packaging.
- Restore the removed focal-pgdg build tooling (debhelper 13, dh-exec, postgresql-common-dev) from apt-archive.postgresql.org.
- Drop the obsolete hurd-iovec patch (merged upstream as of 16.14) and gate on the full quilt series applying cleanly so future drift fails loudly.
- Sign packages with the existing debsigner image (debsigs --sign=maint).

The minor version is parameterized: set PG_UPSTREAM_VERSION (workflow input pg_upstream_version) to build e.g. 16.15; the orig/debian checksums are auto-resolved from the official .dsc unless pinned.

Files:
- dockerfiles/pg16-focal-builder/Dockerfile: focal builder image
- scripts/build_pg16_focal: fetch/verify/assemble/build entrypoint
- .github/workflows/build-pg16-focal.yml: build -> sign -> verify -> upload

### DIFF
--- a/.github/workflows/build-package-test.yml
+++ b/.github/workflows/build-package-test.yml
@@ -13,7 +13,10 @@ env:
   TEST: true
 on:
   push:
-    branches: "**"
+    branches-ignore:
+      # pg16-focal builds PostgreSQL core (see build-pg16-focal.yml); the Citus
+      # extension build/test here is unrelated and should not run on that branch.
+      - pg16-focal
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/build-package-test.yml
+++ b/.github/workflows/build-package-test.yml
@@ -13,10 +13,7 @@ env:
   TEST: true
 on:
   push:
-    branches-ignore:
-      # pg16-focal builds PostgreSQL core (see build-pg16-focal.yml); the Citus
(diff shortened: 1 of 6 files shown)


mtuncer:
we need to build all PG versions we can build not just 16.  try to make it parameterized if possible.
Also think about parameterizing PG minor version so that we won't have to change the script each time PG releases a new minor.

mtuncer:
also, we might want to keep this in its own branch and not merge it into develop, please check with Ibrahim

PR #1185, opened 2026-06-26
Generalize focal PG core build to a parameterized PG13-17 pipeline

kemalbuyukkaya:
## What

A standalone, signed pipeline that rebuilds **PostgreSQL core `.deb` packages for Ubuntu 20.04 (focal)** for every focal-buildable major: **PostgreSQL 13, 14, 15, 16, 17**.

PGDG stopped shipping focal binaries once focal reached EOL standard support (the last official builds were e.g. `16.9-1.pgdg20.04+1`). These five majors still receive newer upstream minors than their last focal build, so we rebuild them ourselves to give focal users current minors with a working, focal-native JIT.

The approach combines the **newer upstream `orig.tar.bz2`** with the **frozen focal-era `debian/` packaging** for each major (its default toolchain is focal's LLVM 10, so the resulting JIT depends on `libllvm10`/`libssl1.1` rather than the `clang-19/llvm-19` that newer packaging requires), then signs with the existing debsigner (`debsigs --sign=maint`).

This is intentionally separate from the Citus extension flow (`citus_package` / `build-package.yml`), which assumes PostgreSQL itself comes from PGDG.

## Changes

**`scripts/build_pg_focal`** — single parameterized build entrypoint
- `PG_MAJOR` selects the major (13–17).
- `PG_UPSTREAM_VERSION` **auto-resolves to the latest minor** from the live PGDG pool (still pinnable) — no edit needed when a new minor is released; orig/debian checksums auto-resolve from the official `.dsc` unless pinned.
- `DEBIAN_BASE` comes from a **static, frozen per-major map** (focal is EOL, so the last focal packaging never changes), overridable via `PG_DEBIAN_BASE`.
- Patch drift is handled declaratively via **`PG_DROP_PATCHES`** (default `hurd-iovec`, a no-op where absent); the quilt-series gate stays as the fail-loud safety net so future drift fails loudly.
- Collects **debug-symbol packages**: `dpkg-buildpackage` emits them as `.ddeb` on Ubuntu, which the old `*.deb` collection dropped. They are now shipped as `.deb` (Debian convention, identical format) so they flow through signing/verification/upload like every other package.

**`dockerfiles/focal-pg-builder/`** — one generic builder image (`focal-pgdg main 13 14 15 16 17`), built once and reused for every major.

**`scripts/smoke_test_focal_debs`** — install smoke test for the built set

The build jobs only prove the packages *exist*, are signed and have unique names — none of that proves they are **installable**. This installs the whole shipped set (including `-dbgsym`) into a stock `ubuntu:20.04`, asserts every package landed at its exact built version (so apt used *our* binaries and did not silently substitute a same-named package), creates and starts a cluster for **every** installed major, and checks `pg_jit_available()` on each — a regression in the frozen-packaging / LLVM-10 strategy that silently disabled JIT would otherwise ship unnoticed.

It also encodes the one external requirement rather than leaving it as tribal knowledge: the server packages need `postgresql-common (>= 252~)`, but focal only ships `214ubuntu0.1`, so the PGDG archive's **`main`** component is enabled to pull `278.pgdg20.04+1`. Only `main` — never the per-major components, which still carry PGDG's own `libpq5`/`libecpg6` for focal and would shadow the packages under test.

**`.github/workflows/build-pg-focal.yml`**
- A `setup` job computes the matrix: a single major via `workflow_dispatch`, or **13–16** by default. PG17 is fully supported but kept out of the default set (not needed up front) — dispatch `pg_major=17` to build it.
- `build-and-sign` matrixes over majors (`fail-fast: false`): build → sign (`citusdata/packaging:debsigner`) → verify `_gpgmaint` → upload a per-major artifact. Output is namespaced per major (`packages/focal/pg<major>/`); the sign step mounts that parent so the signer's `/packages/*/*.deb` glob matches.
- A new **`assemble`** job produces the combined set to publish (see below).
- A new **`install-smoke-test`** job runs `scripts/smoke_test_focal_debs` against that combined set in a clean focal container.
- Runs on every push to `pg-focal` and on `workflow_dispatch`. Top-level `permissions: contents: read` — no job writes back to the repository (artifact up/download use the Actions runtime token, not `GITHUB_TOKEN`).

## Co-installable combined set (`postgresql-all-focal`)

Each major's source builds six **single-instance system libraries** that share one package name but carry a per-major version — `libpq5`, `libpq-dev`, `libpgtypes3`, `libecpg6`, `libecpg-dev`, `libecpg-compat3` (plus their `-dbgsym`). They cannot be co-installed, and the newest copy satisfies every major's `>=` dependency.

The `assemble` job therefore keeps **all per-major packages from every major** but the **shared libraries from the highest major only**, guards against duplicate package names, and uploads a flat, de-duplicated `postgresql-all-focal` artifact. Packages are copied byte-for-byte, preserving their signatures. Per-major artifacts are also retained for traceability.

## Validation

Built **end-to-end for all five majors** (real compiles on focal) — including PG17, which is supported on demand even though it is not in the default matrix. Each produces focal-native packages (`libllvm10`/`libssl1.1`, working JIT); `hurd-iovec` is dropped only on 16/17:

| Major | Auto minor | Frozen base | JIT dep | Patch dropped |
|---|---|---|---|---|
| 13 | 13.23 | 13.21-1.pgdg20.04+1 | libllvm10 | (none) |
| 14 | 14.23 | 14.18-1.pgdg20.04+1 | libllvm10 | (none) |
| 15 | 15.18 | 15.13-1.pgdg20.04+1 | libllvm10 | (none) |
| 16 | 16.14 | 16.9-1.pgdg20.04+1 | libllvm10 | hurd-iovec |
| 17 | 17.10 | 17.5-1.pgdg20.04+1 | libllvm10 | hurd-iovec |

- **dbgsym**: ~11 debug-symbol packages collected per major.
- **Combined set** (validated over all five majors, i.e. a superset of the default 13–16 run): 72 packages, **no duplicate package names**, exactly one 17.10 copy of each shared library, `postgresql-13..17` all present. A default run ships the same shape for 13–16, with the shared libraries taken from PG16.
- **Co-install** (`install-smoke-test`): PG 13–17 install together on a clean focal container with **no conflicts**; every package installs at its exact built version; server binaries report `13.23 / 14.23 / 15.18 / 16.14 / 17.10`; a single `libpq5` 17.10; all 31 dbgsym packages also install; all five clusters start on distinct ports and report `pg_jit_available() = t`.

The signing step itself only runs in CI (needs the deployed `citusdata/packaging:debsigner` image + `PACKAGING_SECRET_KEY`/`PACKAGING_PASSPHRASE`).

## Notes

- 11/12 are intentionally excluded (EOL — no minor newer than their last focal build); 18 never shipped on focal.
- No apt index is generated; the combined `.deb` set is published as-is.
- **Consumers need the PGDG archive's `main` component enabled** (`apt-archive.postgresql.org`, suite `focal-pgdg`): the server packages require `postgresql-common (>= 252~)` and focal only ships `214ubuntu0.1`. Enable `main` only — the per-major components would shadow these packages. This is exercised by `scripts/smoke_test_focal_debs`.

Supersedes #1182.


kemalbuyukkaya:
@microsoft-github-policy-service agree company="Microsoft"

ihalatci (approved):
lgtm

PR #1198, opened 2026-07-10
ci: add gated arm64 (.deb) leg to citus community nightlies

ihalatci:
## What & why

Track 2 of ARM64 support for Citus (citusdata/citus#8612). The Debian-based Citus Docker images install Citus via `apt-get install ... =<pinned-version>` from the packagecloud `citusdata/community(-nightlies)` apt repos. Those repos' **arm64 index is currently empty (0 packages)**, so the Debian images cannot build for `linux/arm64`. This PR starts publishing arm64 `.deb`s by adding a **gated** arm64 leg to the community-nightlies workflow.

## Cross-repo dependencies

- **citusdata/tools#414** replaces the single-architecture Docker SDK build path with buildx and adds gated alpine multi-arch support. It is coordinated with this PR, but its alpine path does **not depend on arm64 `.deb` packages**.
- This PR starts populating the arm64 nightly package index required before the **citusdata/docker** Debian-image workflows can build arm64 variants. Arm64 release legs publishing to `citusdata/community` remain a follow-up and are required for release Docker tags such as `latest` and `postgres-16/17/18`.

## The change (1 file, +25/−1, purely additive)

`.github/workflows/build-citus-community-nightlies.yml`:
- **Gated `arch` matrix dimension**, default OFF:
  `arch: ${{ fromJSON(vars.DEB_BUILD_MULTI_ARCH == 'true' && '["amd64", "arm64"]' || '["amd64"]') }}`.
  When `DEB_BUILD_MULTI_ARCH` is unset/`false` ⇒ `["amd64"]` only ⇒ **the existing amd64 nightly is unchanged**.
- `runs-on: ${{ matrix.arch == 'arm64' && 'ubuntu-24.04-arm' || 'ubuntu-latest' }}` — arm64 legs run on native ARM64 GitHub-hosted runners.
- `exclude: {platform: el/8, arch: arm64}` — `el/8` is an RPM target; arm64 RPM parity is out of scope here.
- Two `if: matrix.arch == 'arm64'` steps (before *Build packages*) that build the arm64 **builder** (`citus/packaging-test:<distro>-all`) and **signer** (`citusdata/packaging:debsigner`) images **natively in-job**, because only amd64 variants of those images are published to Docker Hub today. The build engine's `docker run` has **no `--platform`** flag and prefers a local image, so `citus_package` picks up these native arm64 images automatically; amd64 keeps pulling from Docker Hub.

No changes to the amd64 path, shared scripts (`update_image`, `ci/push_images`), the deb entrypoint, the upload script, or the `citusdata/tools` build engine. Version strings are identical by construction (same `pkgvars` + same engine), satisfying the Docker images' exact-version pins.

## Safety / gating

- **Default OFF.** Without the `DEB_BUILD_MULTI_ARCH` repo variable set to `true`, this is a no-op for the existing pipeline.
- **Publish is still branch-gated.** `upload_to_package_cloud` only publishes when the current branch == the workflow's `main_branch`, so this feature branch never publishes.
- Validated with `actionlint` (clean) and a YAML parse.

## Scope

- **Deb-only.** Produces arm64 `citus` nightlies → `citusdata/community-nightlies` across debian `bullseye`/`bookworm`/`trixie` and ubuntu `jammy`/`noble` (PG majors fan out inside each `-all` build).

## Follow-ups (not in this PR)

- **arm64 release legs** on `all-citus` / `debian-hll` / `debian-topn` (`build-package.yml`, `--build_type release` → `citusdata/community`) — on the critical path for the *release* Docker images (`latest`, `postgres-16/17/18`); also needs the release path's amd64-`jq` hardcode made arch-aware.
- **Publish multi-arch builder/signer images** via buildx (gated) so nightly + release legs skip the per-run in-job image build.
- **RPM arm64 parity** (separate effort).

## Open questions

- Confirm GitHub-hosted `ubuntu-24.04-arm` runners are enabled for the citusdata org (assumed GA/free for public repos).
- Per-nightly cost: the arm64 `-all` builder image is rebuilt each run (cmake + pyenv compile, native arm64) until multi-arch images are published.


.github/workflows/build-citus-community-nightlies.yml
```diff
@@ -74,6 +82,22 @@ jobs:
...
+        run: |
+          export TEST=true
+          export TARGET_PLATFORM="${PLATFORM/\//,}"
+          ./update_image
+        env:
+          PLATFORM: ${{ matrix.platform }}
+
+      - name: Build arm64 debsigner image
+        if: matrix.arch == 'arm64'
+        run: docker build -t citusdata/packaging:debsigner -f dockerfiles/debsigner/Dockerfile .
```
KardelenCeren, line 99: One follow-up suggestion: update the Dockerfile to use a newer base image instead of Xenial, which is outdated and unsupported. Not a blocker for this PR.

KardelenCeren (approved):

PR #1208, opened 2026-08-14
build-pg-focal: refresh the quilt series against the upstream being built

kemalbuyukkaya:
## Problem

14.24/15.19/16.15 add `output_plugin_libraries` to `postgresql.conf.sample` right below the trailing context of the Debian `extension_destdir` hunk. We pair a new orig tarball with the frozen focal-era `debian/`, so the patch stopped matching verbatim and PG14/15/16 failed ([run 31801139919](https://github.com/citusdata/packaging/actions/runs/31801139919)).

The `[4/6]` gate missed it because it was weaker than the real check: `quilt push -a` tolerates fuzz <= 2, while `dpkg-source --before-build` replays the same series in `[6/6]` with `patch -F 0` and refuses it. PG13 was unaffected -- 13.23 doesn't touch that file.

## Fix

`[4/6]` now refreshes the series against the upstream at hand, then gates on zero fuzz:

```sh
quilt push -a --refresh   # apply (fuzz allowed) + rewrite each patch's context
quilt pop -a
quilt push -a --fuzz=0    # what dpkg-source will do in [6/6]
```

Context and line numbers only, no `+`/`-` changes, in PGDG's patch style. A hunk that needed fuzz raises a CI annotation naming the patch and the file, since `--refresh` bakes that placement in and the gate can no longer catch it. A patch that doesn't apply at all points at `PG_DROP_PATCHES`. Frozen packaging plus moving upstream guarantees recurrence, hence a standing fix rather than a per-minor edit.

## Verification

podman locally, plus CI ([run 31805259613](https://github.com/citusdata/packaging/actions/runs/31805259613), all green): refreshed series yields a source tree byte-identical to the fuzzy one; 23 debs per major still linking `libllvm10`, signed; combined 59-deb set installs in stock `ubuntu:20.04` with PG13/14/15/16 online and JIT available.

### DIFF
--- a/scripts/build_pg_focal
+++ b/scripts/build_pg_focal
@@ -93,12 +93,11 @@ ORIG_SHA256="${PG_ORIG_SHA256:-}"
 DEBIAN_SHA256="${PG_DEBIAN_SHA256:-}"
 
 # Patches in the frozen focal debian/ packaging that no longer apply to newer
-# upstream (e.g. merged upstream). Space-separated; each is dropped from the
-# quilt series only if present, so the default is safe for every major:
+# upstream *at all* (e.g. merged upstream). Space-separated; each is dropped from
+# the quilt series only if present, so the default is safe for every major:
 #   * hurd-iovec was merged upstream as of 16.14 (pg_iovec.h guards IOV_MAX),
 #     so it fails to apply on 16.x/17.x but is absent from 13/14/15 packaging.
-# If a future minor introduces new drift, append the offending patch name here
-# (or via the PG_DROP_PATCHES env) -- no other code change required.
+# Context drift needs no entry here -- [4/6] refreshes the series instead.
 PG_DROP_PATCHES="${PG_DROP_PATCHES:-hurd-iovec}"
 
 # Set RUN_TESTS=1 to run the upstream regression suite (slower, needs more deps).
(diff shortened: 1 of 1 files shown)


scripts/build_pg_focal
```diff
@@ -183,11 +182,35 @@ for p in ${PG_DROP_PATCHES}; do
...
+  push_log="${WORK}/quilt-push-pg${PG_MAJOR}.log"
+  quilt push -a --refresh 2>&1 | tee "${push_log}"
+  quilt pop -a > /dev/null
+
+  # A fuzzy hunk was placed on less context than its author wrote: worth a look.
+  if grep -q 'with fuzz' "${push_log}"; then
+    echo "    NOTE: refreshed patches that no longer match ${UPSTREAM} verbatim:"
+    awk '/^Applying patch / { p = $3 }
+         /with fuzz/        { print "      " p ": " $0 }' "${push_log}"
+  fi
```
ihalatci: `--refresh` turns the fuzz case from a hard failure into a log line, and the `--fuzz=0` gate below structurally cannot catch it — refresh has already rewritten the context to match. So a hunk placed on the wrong lines gets baked in as "correct", visible only as a `grep` buried in a long Actions log. Low probability with PGDG's small patches, but it's the one thing this change makes quieter rather than louder.

Suggest surfacing it as a CI annotation and naming the file (`patching file X` always precedes the fuzz line, same as quilt's own `cleanup_patch_output`):

```suggestion
  # A fuzzy hunk was placed on less context than its author wrote: worth a look.
  # --refresh bakes that placement in, so the zero-fuzz gate below can no longer
  # catch it -- surface it as a CI annotation instead of a buried log line.
  if grep -q 'with fuzz' "${push_log}"; then
    echo "::warning::PG${PG_MAJOR}: refreshed patches that no longer match ${UPSTREAM} verbatim"
    awk '/^Applying patch / { p = $3 }
         /^patching file /  { f = substr($0, 15) }
         /with fuzz/        { print "      " p " (" f "): " $0 }' "${push_log}"
  fi
```

`::warning::` is parsed from the container's stdout by the `run` step, so it annotates the run summary; locally it's just one extra line.

Optional, same theme: the `PG_DROP_PATCHES` hint only prints on the zero-fuzz branch, but the likelier failure is a patch that doesn't apply at all (the hurd-iovec class). That path dies via `set -e` with nothing but quilt's raw `does not apply (enforce with -f)`. Wrapping the first push in `if ! ...; then <same hint>; fi` would cost ~3 lines.

Rest of the change looks right to me — I checked quilt 0.66 (focal's version) rather than assuming: `QUILT_REFRESH_ARGS` and `QUILT_DIFF_OPTS` are genuinely honored (`quilt/push:451`, `quilt/refresh:152`), and the second `push -a --fuzz=0` is *not* redundant since the push loop ignores `quilt refresh`'s exit status, so a failed refresh would otherwise pass silently. Good catch on `-F 0` vs. fuzz ≤ 2 being the actual gap.

ihalatci (approved):
approved with minor comment, could be follow up

scripts/build_pg_focal
kemalbuyukkaya: Applied both, thanks -- the point about the gate is right and I'd missed it: `--fuzz=0` validates the *refreshed* series, so by construction it can never see the fuzz. Its value is the other thing you spotted, that `push.in:451` ignores `quilt refresh`'s exit status, so it stays.

One deviation from the snippet. As written, the annotation carries the summary and the patch/file detail lands on following stdout lines, which aren't part of it -- so the run page still says "something drifted" and you open the log anyway. I moved the detail into the annotation, one per fuzzy hunk, which also lets the `grep -q` guard go since awk prints nothing when there's no fuzz:

```
::warning::PG14: extension_destdir needed fuzz against 14.24 in src/backend/utils/misc/postgresql.conf.sample -- Hunk #1 succeeded at 726 with fuzz 2 (offset 1 line).
```

Your `substr($0, 15)` and the `patching file` ordering both check out against the real push logs. Worth flagging that this is now a permanent annotation for 14/15/16 -- focal packaging is frozen, so that drift never goes away. It stays actionable because the text names patch and file, so new drift reads differently, but it is noise rather than an exception.

Optional one applied too. Hard failure now names `PG_DROP_PATCHES`; verified by injecting a non-applying patch, which prints quilt's message, then the hint, then exits 1.

Chasing this turned up a bug in the merged commit that my verification structurally could not see. `patchfns:31` sources `/etc/quilt.quiltrc` *after* the environment, with plain assignments, so focal's rc silently overrode the exports two lines above your comment: `QUILT_REFRESH_ARGS` became `--no-timestamps --backup`, losing `-p ab --no-index`. Refreshed patches came out `Index:`/`src.orig` style instead of PGDG's `a/ b/`, with a `debian/patches/*~` per patch. Harmless -- dpkg-source strips `-p1` either way, which is exactly why the green run hid it -- but the comment claiming PGDG style was false. Fixed with `QUILTRC=/dev/null`. My host harness ran quilt unpacked into a prefix with no rc file to source, so it could never have reproduced it; only the focal image shows it.

PR #1227, opened 2026-09-08
Retire Bullseye from pgautofailover package builds

serhatandic:
## Summary

- retire Debian Bullseye from future pgautofailover package builds after its LTS ended on August 31
- remove Bullseye eligibility from push-triggered and manually dispatched builds
- preserve all other distro and PostgreSQL matrices, build history, and existing packages

## Change

The exact YAML delta was reviewed: one `debian/bullseye` entry was removed from `.github/workflows/build-package.yml`; no other workflow behavior changed.

### DIFF
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -28,7 +28,6 @@ jobs:
           - ol/8
           - debian/stretch
           - debian/buster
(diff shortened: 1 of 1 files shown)


ihalatci:
not operational any more

PR #1228, opened 2026-09-08
Retire Bullseye from pgautofailover enterprise builds

serhatandic:
## Summary

- Remove Debian Bullseye from future pgautofailover enterprise package builds after upstream LTS ended on August 31.
- Leave the other distribution selectors and published package history untouched.

The exact YAML delta was independently reviewed.


ihalatci:
not operational anymore

PR #1235, opened 2026-09-08
Retire Bullseye from enterprise package builds

serhatandic:
## Summary

- Retire Debian Bullseye from future enterprise package build targets following the end of upstream LTS on August 31.
- Leave all other distributions, PostgreSQL versions, build history, and published assets unchanged.

## Change

The exact YAML delta was checked: this removes only the Bullseye entry from the enterprise package build matrix.


ihalatci:
not operation anymore

PR #1236, opened 2026-09-08
Reject Debian Bullseye in public installers

serhatandic:
## Summary

- Reject Debian Bullseye and missing OS metadata (`11`, `11.x`, and `bullseye`) before any APT operations in all four public Debian installer scripts
- Apply the rejection to both automatically detected and explicitly supplied OS inputs
- Remove Bullseye from positive distribution mappings while retaining supported overrides and derivative distributions
- Keep archived Bullseye packages available; this changes installer eligibility only

## Testing

- Detection and differential safety validation were run locally using harnesses outside the repository
- No test framework or standalone harness is added to the public installer source
- No real APT commands or installer deployments were invoked


ihalatci:
no need to block from installer

